PR #14, opened 2015-11-21
Maskbits

sbailey:
This PR puts maskbit code into desiutil.  After it is vetted here, desispec and desitarget can switch to using this.  This code does not actually define the mask bits (that will remain with deispec and desitarget), but it provides convenience wrapper code for manipulating target bits, e.g.:

```
from desiutil.bitmask import BitMask

import yaml
_bitdefs = yaml.load('''
ccdmask:
    - [BAD,       0, "Pre-determined bad pixel (any reason)"]
    - [HOT,       1, "Hot pixel"]
    - [DEAD,      2, "Dead pixel"]
    - [SATURATED, 3, "Saturated pixel from object"]
    - [COSMIC,    4, "Cosmic ray"]
''')
ccdmask = BitMask('ccdmask', _bitdefs)

#- Accessing the mask
ccdmask.COSMIC | ccdmask.SATURATED  #- 2**4 + 2**3
ccdmask.mask('COSMIC')     #- 2**4, same as ccdmask.COSMIC
ccdmask.mask(4)            #- 2**4, same as ccdmask.COSMIC
ccdmask.COSMIC             #- 2**4, same as ccdmask.mask('COSMIC')
ccdmask.bitnum('COSMIC')   #- 4
ccdmask.bitname(4)         #- 'COSMIC'
ccdmask.names()            #- ['BAD', 'HOT', 'DEAD', 'SATURATED', 'COSMIC']
ccdmask.names(3)           #- ['BAD', 'HOT']
ccdmask.comment(0)         #- "Pre-determined bad pixel (any reason)"
ccdmask.comment('COSMIC')  #- "Cosmic ray"    
```


sbailey:
PEP8 and I disagree about comments.  I use '#-' for things that are genuinely comments, to differentiate from real code that was commented out for some reason using '#'.  The Travis pep8 checker is failing on this stylistic choice.  I won't insist that people use my style, but I don't think #- comments should flag red as a failing check.  Can we convert the PEP8 check into an informational build, turn it off, or otherwise distinguish between it and actual tests failing at the top level (i.e. without having to drill down)?

I did appreciate the python3 tests; I learned that python2 will sort a mix of int and str (which I was knowingly doing) but python3 will not, so I had to add some extra workaround code.

@weaverba137 please review the new maskbit code.  100% test coverage :)

weaverba137:
I mainly instituted that as an experiment, just to see how hard it would be.  Not all that hard, actually.  There's no problem granting the occasional exception.  Still, when I ignore the particular error related to comments, E265, then a whole bunch of other errors come up.  I'll clean this up a little bit.

py/desiutil/bitmask.py
```diff
@@ -0,0 +1,162 @@
...
+    ccdmask.mask('COSMIC')     #- 2**4, same as ccdmask.COSMIC
+    ccdmask.mask(4)            #- 2**4, same as ccdmask.COSMIC
+    ccdmask.COSMIC             #- 2**4, same as ccdmask.mask('COSMIC')
+    ccdmask.bitnum('COSMIC')   #- 4
+    ccdmask.bitname(4)         #- 'COSMIC'
+    ccdmask.names()            #- ['BAD', 'HOT', 'DEAD', 'SATURATED', 'COSMIC']
+    ccdmask.names(3)           #- ['BAD', 'HOT']
+    ccdmask.comment(0)         #- "Pre-determined bad pixel (any reason)"
+    ccdmask.comment('COSMIC')  #- "Cosmic ray"
+"""
```
dkirkby: This example might make a good [doctest](https://docs.python.org/2/library/doctest.html).

weaverba137:
Indeed, though it would require a little extra code overhead to run that automatically.

py/desiutil/bitmask.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def __str__(self):
+        return '{:16s} bit {} mask 0x{:X} - {}'.format(
+            self.name, self.bitnum, self.mask, self.comment)
+
+
+#- Class to provide mask bit utility functions
+class BitMask(object):
+    """BitMask object.
+    """
```
dkirkby: It could be useful to have `help(ccdmask)` list the individual bits and their comment strings, rather than a generic `BitMask object` docstring.  One way to accomplish this is to replace `class BitMask` with a function `def createBitMask(name, bitdefs)` that returns
```
type(name, (), class_dict)
```
for an appropriately constructed per-bitmask class dictionary, including a `__doc__` attribute.

weaverba137:
To append to my earlier comment, by default, the pep8 test ignores some errors.  However, by explicitly listing the errors to be ignored, that turns *off* the default set of errors that are ignored.  Thus, it is somewhat easier to clean up the comment strings than to clean up all the errors that would normally be ignored.  So that's what I did.

weaverba137:
If there are no other comments, I think this is ready to merge.

weaverba137:
Hold the presses, just noticed a couple of things...

py/desiutil/bitmask.py
```diff
@@ -0,0 +1,165 @@
...
+    ccdmask.comment('COSMIC')  #  "Cosmic ray"
+"""
+
+
+class _MaskBit(int):
+    """
+    A single mask bit.  Subclasses int to act like an int, but allows the
+    ability to extend with blat.name, blat.comment, blat.mask, blat.bitnum.
+    """
+    def __new__(cls, name, bitnum, comment, extra=dict()):
```
weaverba137: Why are you using extra here instead of `**kwargs`?

py/desiutil/bitmask.py
```diff
@@ -0,0 +1,165 @@
...
+    """BitMask object.
+    """
+    def __init__(self, name, bitdefs):
+        """
+        Args:
+            name : name of this mask, must be key in bitdefs
+            bitdefs : dictionary of different mask bit definitions;
+                each value is a list of [bitname, bitnum, comment]
+
+        Typical users are not expected to create BitMask objects directly.
```
weaverba137: Then how *are* users expected to create BitMask objects?  In the example in the module's docstring, you show a BitMask object being created directly.

py/desiutil/bitmask.py
```diff
@@ -0,0 +1,165 @@
...
+    ability to extend with blat.name, blat.comment, blat.mask, blat.bitnum.
+    """
+    def __new__(cls, name, bitnum, comment, extra=dict()):
+        self = super(_MaskBit, cls).__new__(cls, 2**bitnum)
+        self.name = name
+        self.bitnum = bitnum
+        self.mask = 2**bitnum
+        self.comment = comment
+        for key, value in extra.items():
+            assert key not in (
```
weaverba137: There are  serious problems with this assertion.  First, `int.bitlength` is not actually an attribute of `int`.  It is `int.bit_length`.  Second, this assumes that the set of attributes of `int` will never change.  Finally, it ignores any of the special attributes of `int`.  With this code, one could in principle override `int.__add__`, which I don't think is what was intended.

So, is there any need to handle extra keywords *at all*?  If so, they should be passed (via `**kwargs`) to the base class for handling, then you can use `hasattr()` to test to see whether the attribute already exists, and *then* raise an error if necessary.

py/desiutil/bitmask.py
```diff
@@ -0,0 +1,168 @@
...
+                each value is a list of [bitname, bitnum, comment]
+
+        Typical users are not expected to create BitMask objects directly.
+        """
+        self._bits = dict()
+        self._name = name
+        for x in bitdefs[name]:
+            bitname, bitnum, comment = x[0:3]
+            if len(x) == 4:
+                extra = x[3]
```
weaverba137: OK, now I see where `extra` comes from, but this is still not good, because `_Maskbit` interprets the value of `extra` as a dictionary, but that assumes that `x[3]` is already a dictionary.  How would you write such a thing in a yaml file?

In other words, `BitMask` always assumes that all the parsing of any mask definition is already done.  And it had better be correct!

py/desiutil/bitmask.py
```diff
@@ -0,0 +1,168 @@
...
+
+    #  What to print
+    def __repr__(self):
+        result = list()
+        result.append(self._name + ':')
+        #  return names in sorted order of bitnum
+        bitnums = [x for x in self._bits.keys() if isinstance(x, int)]
+        for bitnum in sorted(bitnums):
+            bit = self._bits[bitnum]
+            result.append('    - [{:16s} {:2d}, "{}"]'.format(
```
weaverba137: Is there a particular reason the name field has to be 16 characters wide?  The problem is when you add in the missing comma, the output looks a bit odd:
```
ccdmask:
    - [BAD             ,  0, "Pre-determined bad pixel (any reason)"]
    - [HOT             ,  1, "Hot pixel"]
    - [DEAD            ,  2, "Dead pixel"]
    - [SATURATED       ,  3, "Saturated pixel from object"]
    - [COSMIC          ,  4, "Cosmic ray"]
```

sbailey:
Thanks for the useful comments.  Catching up on a few of them in a single reply:

this code is intended to replace the code portions of:
  * https://github.com/desihub/desispec/blob/master/py/desispec/maskbits.py'
  * https://github.com/desihub/desitarget/blob/newtargetbits/py/desitarget/targetmask.py

These are nearly identical, awaiting this desiutil code to be ready.  The individual packages would still maintain the bitmask definitions (likely in yaml format), but this desiutil code would provide the common interface object wrapper.  So when the comments say that the typical user is not intended to directly create a bitmask, what I meant was that other packages like desispec and desitarget will use their code to instantiate bitmasks for that package, and normal users will just get the bitmasks pre-made from there.  e.g. "from deispec.bitmask import ccdmask" rather than having the user load a yaml file, pass it to desiutil.bitmask.BitMask, etc.  I'm thinking about how to phrase that more clearly here.

The second case gets its data from https://github.com/desihub/desitarget/blob/newtargetbits/py/desitarget/targetmask.yaml, which demonstrates the use of the "extra" keywords, e.g.
```
targetmask:
    #------ Dark Time Survey: Bits 0-31
    - [LRG,         0, "LRG", {obsconditions: DARK}]
    - [ELG,         1, "ELG", {obsconditions: DARK|GRAY}]
    - [QSO,         2, "QSO", {obsconditions: DARK}]
```
It idea is that in addition to a name, bitnum, and comment, there is optionally an extra dictionary of key/value pairs that get mapped into attributes of the _MaskBit object.  _MaskBit inherits from int so that it can be used as an int, while also extending it with these extra keywords, e.g.
```
targetmask.ELG                # 2**1, acts like an int mask
targetmask.ELG.obsconditions  # 'DARK|GRAY'
```
The code explicitly checks and forbids using extra keywords that would override existing int attributes ('bitlength', 'conjugate', 'real', 'imag', 'numerator', 'denominator').  However, it should also exclude items like __add__.

The original solution was something like
```
targetmask.extra('ELG')['obsconditions']
```
but that just seemed too yucky to remember which things were parentheses vs. square brackets.
It would be good to directly check that the extra dictionary is in fact a dictionary and print a useful exception if it isn't.

@dkirkby suggested enabling help(bitmask).  I like that idea but need a few more minutes to grok the suggested solution.  Note that printing a bitmask will show the info he requested (but only if you know to do that).  The 16-character wide printing for the name was to make it easier to read, but the existing version is not yaml due to missing commas, and it does look a little funny with them in.  A left flushed comma would probably look better (like [desitarget.yaml](https://github.com/desihub/desitarget/blob/newtargetbits/py/desitarget/targetmask.yaml))

weaverba137:
Thanx for the comments.  I'm going to go ahead & implement the `hasattr()` method of determining whether something in `extra` is trying to override an `int` attribute.  Most of the rest is documentation or documentation-related stuff that you should probably handle.

sbailey:
I'm mid-edit with that; if you haven't started yet, then don't start.

weaverba137:
Ah, too late.

sbailey:
OK, Travi's isn't happy about PEP8 check technicalities, and this doesn't have @dkirkby's suggestion of enabling help(ccdmask).  I'd consider both of those optional for this merge (not bad, just optional).  Are we otherwise good to go?

weaverba137:
I can fix those problems then merge, unless you're working on them right now.  I recommend converting @dkirkby's suggestion into an enhancement issue.

dkirkby:
That's fine with me, and then we can discuss whether its worth doing in the new issue (but this is already an "enhancement" issue, so we will need to create a new "enhanced enhancement" label).

sbailey:
I turned off the PEP8 travis tests since the perpetual build failing would become a distraction from actual test failures.  Occasionally checking PEP8 compliance and fixing style stuff is good, but PEP8 compliance isn't an end unto itself.  We should reserve the red "build failing" flag for situations where the tests really are failing and need attention.

PR #16, opened 2015-11-26
Module for fitting 1D arrays

profxj:
Wrappers to numpy polynomial, cheby, etc.
Includes a simple method for iterative fitting with rejection

Includes test suite

### DIFF
--- a/py/desiutil/funcfits.py
+++ b/py/desiutil/funcfits.py
@@ -1,0 +1,207 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+"""
+==================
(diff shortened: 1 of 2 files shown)


sbailey:
Looks good, except a dependency on xpdb slipped in there (hidden under a debug option...).  Please remove that or convert to a vanilla pdb dependency.

Bonus points if you'll also add tests for chebyshev, forceimask, and sigma options, but I'm pleased to see tests included with the PR in the first place!  If easy, let's add those tests but I'm ok with not letting "perfect" 100% coverage be the enemy of the "good" (new features with decent test coverage).

profxj:
xdb removed

tests for cheby,sigma (weights) added

PR #18, opened 2016-01-28
Percentile stat

profxj:
Calculates the percentile bounds of an array

Adds a new stats.py module
Test included

Note: tests are failing on Travis despite passing on my laptop

### DIFF
--- a/py/desiutil/test/test_stats.py
+++ b/py/desiutil/test/test_stats.py
@@ -11,7 +11,7 @@ import pdb
 from desiutil.stats import perc
 
 
-class TestFuncFits(unittest.TestCase):
+class TestStats(unittest.TestCase):
     """Test desiutil.funcfits
     """
 
--- a/py/desiutil/test/test_stats.py
+++ b/py/desiutil/test/test_stats.py
@@ -1,0 +1,33 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+"""Test desiutil.setup.
+"""
+from __future__ import (absolute_import, division,
+                        print_function, unicode_literals)
+# The line above will help with 2to3 support.
+import unittest
+import numpy as np
+import pdb
+from desiutil.stats import perc
+
+
+class TestFuncFits(unittest.TestCase):
+    """Test desiutil.funcfits
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        pass
+
+    @classmethod
+    def tearDownClass(cls):
+        pass
+
+    def test_perc(self):
+        """Test percentile
+        """
+        x = np.linspace(0,np.pi,100)
+        y = np.sin(x)
+        percv = perc(y)
+        np.testing.assert_allclose(percv, np.array([0.22031053278654081,
+                                                   0.96384215855994215]))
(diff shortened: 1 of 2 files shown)


weaverba137:
@profxj I hate to sound grumpy here, but it would be more useful if you could report *why* the Travis build is failing, since you already knew it was failing when you created the PR.

sbailey:
Tests are failing due to missing scipy.  I'm not sure how to tell travis that scipy is a needed, but desispec would be an example.

However, could you use numpy.interp1d instead of scipy.interpolate.interp1d to have a somewhat lighter weight dependency?  Most people probably have both anyway so I'm not sure the difference matters.

weaverba137:
Do we *want* desiutil to depend on both numpy & scipy?

weaverba137:
Also, if we *do* add a dependency on scipy, what version are we targeting?

sbailey:
Consider this implementation instead:

```
import numpy as np
def perc(x, per=68.2):
    return np.percentile(x, [50-per/2.0, 50+per/2.0])
```

Note: given that numpy.percentile uses percentiles in the range [0,100] rather than [0,1], consider following the convention for the range here too.

scipy 0.14 is what we use for desispec tests.  If we can use just numpy for now, that would be slightly preferred, but it seems likely that at some point in the future we'll need a scipy dependency anyway so no big deal if we use it now.

weaverba137:
Hmmm, right now, in the master branch, desiutil doesn't depend on numpy *or* scipy.  The only requirements are (see requirements.txt)

* setuptools
* requests
* pyyaml

These requirements are very simple to install with `pip install -r requirements.txt`.  However, if numpy or scipy are added, pip will try to *compile* numpy (scipy) and that can significantly increase the build time.

weaverba137:
It is *very* strange that the Travis tests passed without installing numpy.

weaverba137:
OK, the tests of the perc function are not actually running because the test class 'TestFuncFits' is being masked by another class of the same name!  This is not good!

py/desiutil/test/test_stats.py
```diff
@@ -0,0 +1,33 @@
...
+from __future__ import (absolute_import, division,
+                        print_function, unicode_literals)
+# The line above will help with 2to3 support.
+import unittest
+import numpy as np
+import pdb
+from desiutil.stats import perc
+
+
+class TestFuncFits(unittest.TestCase):
```
weaverba137: There is already a class called `TestFuncFits` and that class is masking this one.  Change the name to `TestStats`.

py/desiutil/test/test_stats.py
```diff
@@ -0,0 +1,33 @@
...
+    """Test desiutil.funcfits
```
weaverba137: You're testing desiutil.stats, not desiutil.funcfits.

py/desiutil/test/test_stats.py
```diff
@@ -0,0 +1,33 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+"""Test desiutil.setup.
```
weaverba137: You're testing desiutil.stats not desiutil.setup.

profxj:
ps.  I don't think the duplicate name was colliding, i.e. 
both tests were running (each is in its own file, so I suspect
have unique names).

pps.  Tests passing without having done anything about numpy.
Of course, they were before (see the real TestFuncFits)  ;)

PR #23, opened 2016-03-14
add code to download data not bundled with a package

### LINKED ISSUE #22: Allow desiInstall to grab data files before installation
Some packages need to grab external data to be a "complete" package.  Doing this during the `python setup.py install` stage is tricky (it only knows about the Python code), but since desiInstall copies the complete package to the final install directory *before* performing `python setup.py install`, one could just grab the data into the *source* code and *then* perform the copy.

This is related to desihub/desimodel#11.

weaverba137:
Fixes #22.

desiInstall will search for a script called `etc/<product>_data.sh` (*e.g.* `etc/desimodel_data.sh`) and execute it prior to copying the downloaded code to `INSTALL_DIR`.  The script can assume the existence of a few environment variables that might be useful for 'positioning' the download:

* `WORKING_DIR`
* `<PRODUCT>_VERSION` (*e.g.* `DESIMODEL_VERSION`)


sbailey:
Clarifying the documentation: if the etc/product_data.sh script is detected, the installation will run it.  The intended use case is that that script will download other necessary data, but the script could be used for other purposes as well for configuring things that for whatever reason you don't want to do with setup.py itself.  Correct?

Otherwise, this looks fine; feel free to merge after tests have finished running.

weaverba137:
No, the script is *only* intended to download the absolutely necessary data, and it should only be used with desiInstall or Travis.  It should not be used for anything else.

Having it do anything else is just scope creep.

See the extra-data branch in desimodel for an example.

weaverba137:
To further clarify:  the scope of this PR is to deal with the situation where necessary data *is not* bundled with the code.  There are other, better ways for locating and manipulating data that *is* bundled with the code.

PR #25, opened 2016-03-24
fix module processing problem

### LINKED ISSUE #24: desiutil.modules.configure_module doesn't handle non-DESI products exactly correctly.
Packages like speclite and specsim follow the normal Python convention of having the top-level directory of the Python package in the top-level directory of the repository as a whole.  In contrast, DESI uses a different convention of placing the top-level directory of a Python package inside a `py/` directory which is then placed in the top-level directory of the hierarchy.

`desiutil.modules.configure_module()` still only supports the DESI convention, so speclite and specsim will not have the PYTHONPATH set correctly in their Module files.

weaverba137:
This PR:

* Fixes #24.
* Allows the NERSC module root directory to be overridden in a configuration file.


weaverba137:
Latest changes should also fix #26.  I'm still looking closely at this, so don't merge yet.

PR #27, opened 2016-03-24
Generate fit_dict for funcfits

profxj:
Simple algorithm to generate a fit_dict for funcfits.

I believe code like this existed previously (it is called
by desi_bootcalib) but I can't find it..

Includes a test.


sbailey:
This is out of context for me.  What is a fit_dict?  Why is this needed / how is it used?  Please update docstring with what the params are and how the returned dictionary should be used.

e.g. from the tests I see that func='legendre' is a valid choice, but is 'chebyshev'?  Others?  User defined functions?  From the test I don't see why 10 coefficients for a order=5 polynomial should work, etc.  Thanks.

profxj:
Have added docs.

This is used when one needs to generate the output from
a fit without actually performing the fit.  Presently, only applied
in desi_bootcalib when running only on an Arc frame.

weaverba137:
Why is the fit_dict branch still present?

PR #30, opened 2016-04-10
Func Fits: Weights and bug fix

profxj:
Enables use of weights
Fixes a bug in final fit


weaverba137:
In the future, can you please insist that all PRs on desiutil include an update to changes.rst prior to merging?  That or only let me do all PR merges.

sbailey:
Thanks for the reminder.  "git log --oneline" can help retroactively figure out what should have been in changes.rst, but it is less human-readable than a curated changes.rst .  Directly editing docs/changes.rst via the github can be a lightweight way of adding a comment if one initially forgot.

weaverba137:
In this case `git log --oneline` gives "weights and mask".  Not very enlightening.

PR #31, opened 2016-04-27
Method to 'yamilify' a Python object

profxj:
Method to convert a Python object (often a dict) into a yaml-valid 
object.  Mainly to be used for writing to yaml.

Am pretty sure this is Python3 compatible (passes Travis and some
of my own tests), but I am not fully confident of all cases.

Creates a new module (io.py)
Updated changes.rst
Added unittest

### DIFF
--- a/py/desiutil/io.py
+++ b/py/desiutil/io.py
@@ -1,0 +1,74 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+"""
+==================
+desiutils.io
+==================
+
+Module for I/O related code
+"""
+from __future__ import (print_function, absolute_import, division,
+                        unicode_literals)
+import numpy as np
+
+
+def yamlify(obj, debug=False):
+    """ Recursively process an object so it can be serialised for yaml
+    Based on jsonify in linetools
+
+    Note: All keys in dicts are converted to str type
+
+    Parameters
+    ----------
+    obj : any object
+    debug : bool, optional
+
+    Returns
+    -------
+    obj - the same obj is yaml_friendly format
+      (arrays turned to lists, np.int64 converted to int, np.float64 to float,
+      basestring to str, etc.).
+
+    """
+    if isinstance(obj, np.float64):
+        obj = float(obj)
+    elif isinstance(obj, np.float32):
+        obj = float(obj)
+    elif isinstance(obj, np.int32):
+        obj = int(obj)
+    elif isinstance(obj, np.int64):
+        obj = int(obj)
+    elif isinstance(obj, np.int16):
+        obj = int(obj)
+    elif isinstance(obj, np.bool_):
+        obj = bool(obj)
+    elif isinstance(obj, (np.string_,basestring)):
+        obj = str(obj)
+    #elif isinstance(obj, Quantity):
+    #    obj = dict(value=obj.value, unit=obj.unit.to_string())
+    #elif isinstance(obj, np.ndarray):  # Must come after Quantity
+    #    obj = obj.tolist()
+    elif isinstance(obj, dict):
+        # First convert keys
+        obj = {str(key):value for key,value in obj.items()}
+        # Now recursive
+        for key, value in obj.items():
+            obj[key] = yamlify(value, debug=debug)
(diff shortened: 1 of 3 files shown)


sbailey:
Looks good to me; thanks for including tests and a changes.rst bullet.

FYI, isinstance also accepts a tuple of types to check which could compactify the logic of input types -> output types, e.g.
```
elif isinstance(obj, (np.int16, np.int32, np.int64)):
        obj = int(obj)
```
Update: I just noticed that you did this for strings but not integers, but I'll leave my comment in place anyway.

I'll give Ben a chance to comment, but if we don't hear back from him soon I think it is fine to merge this when you need it.

profxj:
Have consolidated those.

sbailey:
Why is the np.ndarray case commented out?  It seems likely we would want that someday.

py/desiutil/io.py
```diff
@@ -0,0 +1,72 @@
...
+    Note: All keys in dicts are converted to str type
+
+    Parameters
+    ----------
+    obj : any object
+    debug : bool, optional
+
+    Returns
+    -------
+    obj - the same obj is yaml_friendly format
```
weaverba137: What does "the same obj is yaml_friendly format" mean?  This would be better as:
```
Returns
-------
object
    An object suitable for yaml serialization.  For example :class:`numpy.ndarray` is converted to :class:`list`, :class:`numpy.int64` is converted to :class:`int`, etc.

File: py/desiutil/test/test_io.py [hunk 2]
```diff
@@ -0,0 +1,45 @@
...
+"""Test desiutil.io.
+"""
+from __future__ import (absolute_import, division,
+                        print_function, unicode_literals)
+# The line above will help with 2to3 support.
+import unittest
+import sys
+import numpy as np
+from warnings import catch_warnings, simplefilter
+import pdb
```
weaverba137: Why is pdb being imported here?

py/desiutil/test/test_io.py
```diff
@@ -0,0 +1,45 @@
...
+# -*- coding: utf-8 -*-
+"""Test desiutil.io.
+"""
+from __future__ import (absolute_import, division,
+                        print_function, unicode_literals)
+# The line above will help with 2to3 support.
+import unittest
+import sys
+import numpy as np
+from warnings import catch_warnings, simplefilter
```
weaverba137: You're importing from warnings but not actually using these as far as I can tell.

py/desiutil/test/test_io.py
```diff
@@ -0,0 +1,45 @@
...
+    def tearDownClass(cls):
+        pass
+
+    def test_yamlify(self):
+        """Test yamlify
+        """
+        fdict = {'name':'test', 'num':np.int32(3), 'flt32':np.float32(3.), 'flt64':np.float64(2.),
+                     'num2':np.int64(4), 'bool':np.bool(True), 'lst':['tst2', np.int16(2)],
+                     'tup':(1,3), 'dct':{'a':'tst3', 'b':np.float32(6.)}}
+        assert isinstance(fdict,dict)
```
weaverba137: You should use the `assert*` methods of `unittest.TestCase` objects to be fully compatible with `unittest`.  This line should be `self.assertIsInstance(fdict, dict).  Although I'm not exactly sure why you need to test that since it is manifestly a dictionary.

weaverba137:
I've added a bunch of line comments.  Please address those.  They're all fairly minor.

py/desiutil/io.py
profxj: edited.  And I've turned on the ndarray conversion.
Wasn't sure it was necessary for yaml (this was a json port),
but I see now it is.

py/desiutil/test/test_io.py
profxj: For development.  Commented out.

py/desiutil/test/test_io.py
profxj: Removed.

py/desiutil/test/test_io.py
profxj: Changed.

profxj:
All addressed,  I think.

Please merge if you approve.

PR #36, opened 2016-06-09
add desiutil.depend module

sbailey:
This PR adds a `desiutil.depend` module with utilities for working with code dependencies stored in FITS headers.  The code name and the code version are stored in pairs of keywords, similar to how table columns are defined.  e.g.
```
DEPNAM00 = 'numpy'
DEPVER00 = '1.11'
DEPNAM01 = 'desiutil'
DEPVER01 = '1.4.1'
```
The functions and Dependencies class provide convenience wrappers to loop over they keywords looking for a particular dependency and adding a new dependency version to next available DEPNAMnn/DEPVERnn.

Examples:
```
>>> import desiutil
>>> from desiutil import depend
>>> import astropy
>>> from astropy.io import fits
>>> 
>>> hdr = fits.Header()
>>> depend.setdep(hdr, 'desiutil', desiutil.__version__)
>>> depend.setdep(hdr, 'astropy', astropy.__version__)
>>> depend.getdep(hdr, 'desiutil')
'1.4.1.dev316'
>>> depend.hasdep(hdr, 'astropy')
True
>>> hdr
DEPNAM00= 'desiutil'
DEPVER00= '1.4.1.dev316'
DEPNAM01= 'astropy '
DEPVER01= '1.1.1   '
```
There is also an object wrapper that gives a header dict-like semantics to update or view dependencies.  This directly updates the input header object so that it can be used in subsequent I/O
```
>>> codever = depend.Dependencies(hdr)
>>> codever['blat'] = '1.2'
>>> codever['foo'] = '3.4'
>>> for codename, version in codever.items():
...     print(codename, version)
... 
('desiutil', '1.4.1.dev316')
('astropy', u'1.1.1')
('blat', '1.2')
('foo', '3.4')
```

This includes tests and an update to changes.rst .  After this is merged I'd like to tag 1.5.0 so that this can be used in desispec.

### DIFF
--- a/py/desiutil/depend.py
+++ b/py/desiutil/depend.py
@@ -113,6 +113,19 @@ def hasdep(header, name):
     except KeyError:
         return False
 
+def iterdep(header):
+    '''Returns iterator over (codename, version) tuples'''
+    for i in range(100):
+        namekey = 'DEPNAM{:02d}'.format(i)
+        verkey = 'DEPVER{:02d}'.format(i)
+        if namekey not in header and i==0: continue
+        if namekey in header:
+            yield (header[namekey], header[verkey])
+        else:
+            raise StopIteration
+
+    raise StopIteration
+
 class Dependencies(object):
     """Dictionary-like object to track dependencies.
 
@@ -144,18 +157,9 @@ class Dependencies(object):
 
     def __iter__(self):
         '''Returns iterator over name.'''
-        for i in range(100):
-            namekey = 'DEPNAM{:02d}'.format(i)
-            verkey = 'DEPVER{:02d}'.format(i)
-            if namekey not in self.header and i==0: continue
-            if namekey in self.header:
-                yield self.header[namekey]
-            else:
-                raise StopIteration
-
-        raise StopIteration
+        for name, version in iterdep(self.header):
+            yield name
 
     def items(self):
         '''Returns iterator over (name, version).'''
-        for name in self:
-            yield name, self[name]
+        return iterdep(self.header)
--- a/py/desiutil/depend.py
+++ b/py/desiutil/depend.py
@@ -1,15 +1,19 @@
-#!/usr/bin/env python
-
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
 """
+===============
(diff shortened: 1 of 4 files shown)


py/desiutil/depend.py
```diff
@@ -0,0 +1,152 @@
...
+        header : dict-like object to store dependencies
+        name : code name string
+        version : version string
+    '''
+    for i in range(100):
+        namekey = 'DEPNAM{:02d}'.format(i)
+        verkey = 'DEPVER{:02d}'.format(i)
+        if namekey in header:
+            if header[namekey] == name:
+                header[namekey] = version
```
weaverba137: I'm fixing this myself on the next commit, but I just wanted to point out that this wouldn't have worked.

py/desiutil/depend.py
```diff
@@ -0,0 +1,152 @@
...
+    def __init__(self, header=None):
+        '''Initialize Dependencies with dict-like header object
+        
+        if header is None, use a empty OrderedDict
+        '''
+        if header is None:
+            from collections import OrderedDict
+            self.header = OrderedDict()
+        else:
+            self.header = header
```
weaverba137: I'm not sure I like this.  The problem is that this creates a reference to header that can continue to exist even if header goes out of scope, as long as the Dependencies object continues to exist.  We already have to deal with FITS objects having dangling references, & since this is intended to be used with FITS headers, we're potentially creating yet another way of preventing FITS objects from being garbage-collected.

sbailey:
Thanks for the bug fix; I had already caught that and fixed it including a test, but had forgotten to commit before pushing...  I merged my changes with yours and committed and re-pushed.

I also added an `iterdep` function to iterate over (codename,version) pairs so that all features are possible with both functions and the object wrapper.

Dependencies keeping a reference to original header: Good issue to raise but I think this is ok.  We do similar to this throughout the desispec code, with many classes like Frame and FiberFlat keeping references to headers and data that come from a FITS file.  I imagine that Dependencies objects will be shorter lived, used as syntactic sugar within a function, while the data wrapper classes like Frame are the longer lived objects that will drive the garbage collection.

weaverba137:
Tag when ready.

PR #37, opened 2016-06-10
add add_dependencies()

sbailey:
Adds `desiutil.depend.add_dependencies(header, module_names=None)` function that will loop over known possible DESI dependencies and add them to the header only if they have already been imported by the code.  This assumes that if the module is loaded, it was loaded for a good reason and should be listed as a dependency.  Conversely, if a module wasn't loaded by the time you got far enough to write a file then it wasn't actually a dependency for that file.  This saves individual packages or I/O routines from having to maintain their own list of what might have been a dependency.  `module_names` allows the user to override with a different list of module names to check.  This does not go so far as trying to automagically detect every loaded module that isn't part of the standard library.

The intension is that every `io.write*` function in `desisim`, `desispec`, and `desitarget` should call this to update the primary HDU header before writing it out.

The current list of modules checked is:
```
        module_names = [
            'numpy', 'scipy', 'astropy', 'yaml', 'matplotlib',
            'requests', 'fitsio', 'h5py', 'mpi4py', 'psycopg2',
            'desiutil', 'desispec', 'desitarget', 'desimodel', 'desisim',
            'redmonster',
            'specter', 'speclite', 'specsim',
            ]
```
Some of these are not yet dependencies but might be in the future; the "is it loaded" check is easy and fast and is ok if the module isn't even installed, so we can err on the side of future completeness.

Are there others that should be included?

### DIFF
--- a/py/desiutil/depend.py
+++ b/py/desiutil/depend.py
@@ -126,6 +126,14 @@ def iterdep(header):
 
     raise StopIteration
 
+#- default possible dependencies to check in add_dependencies()
+possible_dependencies = [
+    'numpy', 'scipy', 'astropy', 'yaml', 'matplotlib',
+    'requests', 'fitsio', 'h5py', 'mpi4py', 'psycopg2',
+    'desiutil', 'desispec', 'desitarget', 'desimodel', 'desisim',
+    'redmonster', 'specter', 'speclite', 'specsim',
+    ]
+
 def add_dependencies(header, module_names=None):
     '''Adds DEPNAMnn, DEPVERnn keywords to header for imported modules
     
@@ -134,7 +142,7 @@ def add_dependencies(header, module_names=None):
     
     Options:
         module_names : list of module names to check
-            if None, check a list of known potential DESI dependencies
+            if None, checks desiutil.depend.possible_dependencies
         
     Only adds the dependency keywords if the module has already been
     previously loaded in this python session.  Uses module.__version__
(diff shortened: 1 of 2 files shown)


py/desiutil/depend.py
```diff
@@ -126,6 +126,48 @@ def iterdep(header):
...
+        
+
+    #- Set version strings only for modules that have already been loaded
+    for module in module_names:
+        if module in sys.modules:
+            #- already loaded, but we need a reference to the module object
+            x = importlib.import_module(module)
+            try:
+                version = x.__version__
+            except AttributeError:
```
weaverba137: Please add a test that covers this exception.

py/desiutil/depend.py
```diff
@@ -126,6 +126,48 @@ def iterdep(header):
...
+    previously loaded in this python session.  Uses module.__version__
+    if available, otherwise "unknown (/path/to/module/)".
+    '''
+    import sys
+    import importlib
+    
+    setdep(header, 'python', sys.version.replace('\n', ' '))
+    
+    if module_names is None:
+        module_names = [
```
weaverba137: I don't think we want to "bury" this list inside this function.  It should be a module-level variable or read from a configuration file.

sbailey:
I pulled the default list of modules to check into a module level variable:
```
possible_dependencies = [
    'numpy', 'scipy', 'astropy', 'yaml', 'matplotlib',
    'requests', 'fitsio', 'h5py', 'mpi4py', 'psycopg2',
    'desiutil', 'desispec', 'desitarget', 'desimodel', 'desisim',
    'redmonster', 'specter', 'speclite', 'specsim',
    ]
```
I added a test for the case of a module that doesn't have \_\_version\_\_ and expanded the cases, though it still doesn't check the exact case of redmonster: a non standard library module that doesn't set \_\_version\_\_ (but that case does work, as checked by hand since I do have redmonster installed).

dkirkby:
It would be good to assert, or at least issue a warning, if this runtime dependency detection identifies anything not already listed in the package metadata used by setup.py.

Are you assuming that __version__ is sufficient to capture the provenance of an output?  That seems quite fragile, especially if you expect reproducibility.  Git commit hashes would be much better, although they are pain to obtain in general (but perhaps this only needs to work in the controlled production environment?)

weaverba137:
@sbailey, thank you for adding the test, but my original purpose in asking that (I know, I didn't spell it out) was that after the previous PR, the coverage of `desiutil.depend` was at 100%, & I was hoping that once it was at 100% it would stay at 100%.

sbailey:
@weaverba137 100% coverage is a nice goal, but let's not make it a fetish.  The one line of `depend.py` code that isn't tested by travis is tested on your laptop if you have redmonster installed.  We've got more important pieces of code with way less coverage, and completely missing code still to write.

@dkirkby autodetecting discrepancies with setup.py / requirements.txt would be nice but I'd be inclined to implement that as a separate test within each package or a separate tool for determining what should be in requirements.txt etc.

This header keyword version provenance is complimented by the pipelining infrastructure that Ted set up that includes in the pipeline output a script defining how to configure the environment.  That is the more directly useful way of achieving reproducibility.  Including version strings in the headers is a belt-and-suspenders backup and let's the files be semi-standalone for knowing versions, even if they aren't structurally guaranteed to be accurate.

Bottom line: Let's not hold up this step just because more could be done.

PR #39, opened 2016-07-19
Combine dicts

profxj:
Simple method to combine 2 dicts
Useful for QA if nothing else.

Includes a test

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -12,6 +12,7 @@ Change Log
 
 * Fixed some import statements so documentation will build on readthedocs.
 * depend.add_dependencies() to add DEPNAM/DEPVER for common DESI dependencies
+* Added combineDicts method
 
 1.5.0 (2016-06-09)
 ------------------
--- a/py/desiutil/io.py
+++ b/py/desiutil/io.py
@@ -24,13 +24,13 @@ def combineDicts(dictionary1, dictionary2):
         output : dict
     """
     output = {}
-    for item, value in dictionary1.iteritems():
+    for item, value in dictionary1.items():
         if dictionary2.has_key(item):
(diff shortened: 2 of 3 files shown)


sbailey:
minor style: should this be combine_dicts instead of combine_dicts?

The `dictionary2.pop(item)` means that dictionary2 can be modified by this call, which I don't think was intended.

Please document a bit more what is meant by "respecting common keys" and provide more tests showing that it does that.  e.g. exactly what is supposed to happen if both dictionaries have the same key?  In the current implementation, if they share a key whose value is a dictionary, the values get merged, but if they share a key whose value is a scalar, then dictionary2 wins.  And if they share a key where one has a scalar value and the other has a dictionary, then it depends upon the order for whether dictionary2 wins or whether it generates an AttributeError.  Are all those cases intended?

sbailey:
Thinking about this more, I suspect that the intended use was for cases where there never would be a leaf-level collision when merging the two dictionaries, e.g., merging two QA frames from the same exposure or merging two exposures from the same night.  If that is the case, I suggest making that explicit: if `x` and `y` both have key `a`, then `x[a]` and `b[a]` must both be dictionaries to recursively merge, or else raise an exception.  If the code ever finds itself in a situation where it is choosing between `x` and `y` instead of merging `x` and `y`, then raise an exception instead of silently choosing one over the other (and document that leaf level conflicts aren't allowed).

Of course, if you do anticipate supporting leaf-level conflicts, then document what the merging code will do in that case...

profxj:
Addressed these useful comments.

profxj:
ok, I think I have it.

sbailey:
I added more tests just to be sure and I didn't find any more bugs.  :)  I'll merge as soon as the Travis tests have finished.

PR #40, opened 2016-07-29
Plot Slices 

profxj:
Generates a scatter plot and histogram in slices using
a modified version of code written originally by D. Kirkby.

Here is a snapshot of the test:

test


profxj:
Merging (as per SJB instructions).

PR #41, opened 2016-09-01
Encoding

sbailey:
This PR adds `desiutil.io.encode_table()` and `decode_table()` to convert table columns between unicode and bytes, useful for converting python 3 strings into bytes prior to writing out to FITS of HDF5.  `decode_table` includes the option to decode to the native python string type, i.e. decode for python 3 but not for python 2.

`encode_table` will be needed for desihub/desispec#266 .  We don't actually need `decode_table` yet because `astropy.io.fits` and `astropy.table.Table` try to do that for you, but it could come in handy if we switch to using fitsio or h5py.

This does introduce a dependence upon astropy, which is new but I think OK.  I put this into requirements.txt so that (hopefully) travis tests will install that.  In the code, astropy is only imported if `encode_table` or `decode_table` are actually called.  There may be a better way to install it for tests without foisting it upon everyone via requirements.txt; @weaverba137 can comment if this should be done differently.

### DIFF
--- a/py/desiutil/io.py
+++ b/py/desiutil/io.py
@@ -115,6 +115,31 @@ def _dtype_size(dtype):
     i = dtype.str.find(dtype.kind)
     return int(dtype.str[i+1:])
 
+def _pick_encoding(table, encoding):
+    '''
+    pick which encoding to use; giving warning if options are in conflict
+
+    Args:
+        table : astropy Table object
+        encoding : (str) encoding to use; None to use table.meta['ENCODING']
+
+    Note: `encoding` trumps `table.meta['ENCODING']`
+    '''
+    if encoding is None:
+        if 'ENCODING' in table.meta:
+            encoding = table.meta['ENCODING']
+        else:
+            raise UnicodeError('No encoding given as argument or in table metadata')
+    elif 'ENCODING' in table.meta and table.meta['ENCODING'] != encoding:
+        import warnings
+        message = """\
+data.metadata['ENCODING']=='{}' does not match option '{}';
+use encoding=None to use data.metadata['ENCODING'] instead""".format(\
+            table.meta['ENCODING'], encoding)
+        warnings.warn(message)
+
+    return encoding
+
 def encode_table(data, encoding='ascii'):
     '''
     Encode unicode strings in a table into bytes using numpy.char.encode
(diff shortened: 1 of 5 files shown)


py/desiutil/io.py
```diff
@@ -106,3 +106,87 @@ def yamlify(obj, debug=False):
...
     return obj
+
+def _dtype_size(dtype):
+    '''
+    Parse dtype like '<Un' into int(n)
+    Note that this is different from dtype.itemsize, which is number of bytes
+    '''
+    i = dtype.str.find(dtype.kind)
+    n = int(dtype.str[i+1:])
+    return int(n)
```
weaverba137: Why do you `int()` twice?

py/desiutil/io.py
```diff
@@ -106,3 +106,87 @@ def yamlify(obj, debug=False):
...
+        UnicodeError if no encoding is given as argument or in table metadata
+
+    Note: `encoding` option overides data.meta['ENCODING'];
+        use encoding=None to use data.meta['ENCODING']
+    '''
+    from astropy.table import Table
+    import numpy as np
+    table = Table(data, copy=False)
+
+    if encoding is None:
```
weaverba137: In order for this to be true, you would have to deliberately set `encoding=None` in the arguments to the function.

py/desiutil/test/test_io.py
```diff
@@ -8,8 +8,9 @@
 import unittest
 import sys
 import numpy as np
+from astropy.table import Table
 #import pdb
-from desiutil.io import yamlify, combine_dicts
+from desiutil import io
```
weaverba137: This is dangerous, because it masks the Python 3 io library. https://docs.python.org/3/library/io.html

weaverba137:
If Travis builds start taking a really long time, that will be a sign that pip is compiling astropy, rather than installing it from a pre-built binary.  Normally it will do the latter, but unusual combinations of versions could trigger a compile.

py/desiutil/io.py
weaverba137: `return int(dtype.str[i+1:])` would be simpler.

weaverba137:
PS, to speed up Travis tests a little bit, I don't think we need Python 3.3 tests at this point.

weaverba137:
One more thing: don't forget to update the changes.rst file.  I got a gentle reminder about that from the astropy folks today, so I'm passing that on.

sbailey:
I did some cleanup as suggested by Ben and dropped the python 3.3 tests.  Individual travis tests took <4 min, but their combination took almost 30 min to make it through the Travis queue.  Oddly, build_sphinx took the longest.

By default, `encoding='ascii'`, and this overrides anything in the data file since an ENCODING='ascii' (or otherwise) header is non-standard usage and I didn't want to rely upon that keyword being there.  To force the code to look for an ENCODING keyword, use `encoding=None`.  I added check for the user specifying an encoding that is different from what is in the file (if specified) and issue warning, but don't raise an error.

I think we don't want to rely upon ENCODING in the file as the default (i.e. keep `encoding='ascii'` as default, while allowing `encoding=None` to fall back to the file).  But perhaps inconsistent encoding specifications should raise an error instead of just a warning.

weaverba137:
Looks good, just add a note to the changes.rst file...

sbailey:
Updated changes.rst and worked around another FITS_rec / Table bug with unicode vs. bytes...  Thanks for the suggestions; will merge once tests have completed.

PR #45, opened 2016-09-27
Mask names uint64

sbailey:
This PR
* fixes desihub/desitarget#79 that reported that `BitMask.names(x)` didn't work when `x` was a `np.uint64` due to numpy/numpy#2955 .
* fixes a `2**n` vs. `n**2` bug that makes `BitMask.names` much faster :)
* absolute instead of relative import of `desiutil.bitmask` within unit test so that it can be run standalone


py/desiutil/test/test_bitmask.py
```diff
@@ -6,8 +6,9 @@
                         print_function, unicode_literals)
 # The line above will help with 2to3 support.
 import unittest
-from ..bitmask import BitMask, _MaskBit
+from desiutil.bitmask import BitMask, _MaskBit
```
weaverba137: I don't understand what you mean by "making a test run standalone".  Also, this is not actually an absolute import.  This is just poor import style.

py/desiutil/test/test_bitmask.py
```diff
@@ -105,6 +106,29 @@ def test_str(self):
             self.assertEqual(bitmask[name].bitnum, self.ccdmask[name].bitnum)
             self.assertEqual(bitmask[name]._extra, self.ccdmask[name]._extra)
 
+    def test_uint64(self):
+        _bitdefs = dict(ccdmask=list())
+        _bitdefs['ccdmask'].append( ['BAD',   0, "badness"] )
+        _bitdefs['ccdmask'].append( ['HOT',   1, "hothot"] )
+        _bitdefs['ccdmask'].append( ['TEST', 16, "testing"] )
+
```
weaverba137: I recommend actually adding a bit mask with a values larger than `2**16`, including `2**31`, `2**32` and `2**63`.

py/desiutil/test/test_bitmask.py
sbailey: With the original relative import syntax, you cannot run
```
[desiutil] $ python py/desiutil/test/test_bitmask.py
Traceback (most recent call last):
  File "py/desiutil/test/test_bitmask.py", line 9, in <module>
    from ..bitmask import BitMask, _MaskBit
ValueError: Attempted relative import in non-package
```
With the old-style fully qualified package path syntax, you can:
```
[desiutil] $ python py/desiutil/test/test_bitmask.py
.......
----------------------------------------------------------------------
Ran 7 tests in 0.011s

OK
```
This is a handy trick when debugging a single failing test since you don't have to keep rerunning every passing test via `python setup.py test`.  I always re-run that at the end to make sure I didn't accidentally break something else, but being able to run just a single test file is useful.  This is even more useful for packages like desisim where the full suite of tests can take a few minutes to run.

If the old-style syntax actually creates a problem in this case, let me know.

py/desiutil/test/test_bitmask.py
sbailey: Good suggestion.  Implementing that revealed that trying to define the highest bit doesn't work under python 2.7, since the _MaskBit object inherits from python int, which can't store 2**63 in python 2.7 but it can under python 3.5 (where it gracefully becomes a long under the hood).  I added a commented out unittest as a placeholder for debugging that but I suggest that this gets spun off to a different issue and PR and not be a blocking factor for this one.

When we really deprecate python 2.7 this becomes a moot point — mask bits with the highest bit set do work under python 3.5.

py/desiutil/test/test_bitmask.py
weaverba137: OK, I can understand the desire to run individual tests, but there are other ways to do this.

First let me clear up a misconception:  "absolute import" is a bad name.  It should be called "fully-qualified import".  In fact both `from ..bitmask import` and `from desiutil.bitmask import` are fully-qualified, and therefore "absolute" imports.  If you were in the top-level `__init__.py` file for desiutil and wrote `import bitmask`, that would *not* be fully-qualified and therefore not "absolute".

Now here's the dangerous part.  With `from ..bitmask import`, there is absolutely no ambiguity about which module you want: "I want the module in the directory above."  With `from desiutil.bitmask import` it's a bit tricky.  If the module in the directory above is the only possibility, or if it *happens to be first* in the `sys.path`, then your test should be testing against the correct module.  However, because desiutil is so heavily used, it is entirely possible that you could be running tests on desiutil while there is another version of desiutil in your environment.

Does that all make sense?

py/desiutil/test/test_bitmask.py
weaverba137: Yes, you will certainly have problems in Python 2.  Instead of commenting out the test, I would add it as a skipped test or an expected failure under Python 2.  Technically under Python 3.5, the integers are not "gracefully becoming long".  There is no `long()` in Python 3.  All integers in Python 3 are what used to be called long in Python 2.

py/desiutil/test/test_bitmask.py
sbailey: Thanks for the explanation.  I verified that this is a real world problem by trying to test a fresh git clone that wasn't in my PYTHONPATH while an older reference version was.  It picked up the wrong version as you said.

I reverted the import syntax, but I'd appreciate hearing the "other ways to do this".  Do you know a convenient one-liner for running only a single test file instead of all of them, while using the `from ..bitmask import BitMask` syntax?

py/desiutil/test/test_bitmask.py
sbailey: I changed the decorator to skip the test under python 2.x.  I had originally flagged it as expectedFailure, but then it unexpectedly succeeded under python 3.5.  I don't see an obvious way to conditionally flag something as expectedFailure.

py/desiutil/test/test_bitmask.py
weaverba137: I think I can come up with an example individual test relatively quickly.

py/desiutil/test/test_bitmask.py
weaverba137: That's fine.

py/desiutil/test/test_bitmask.py
weaverba137: You will now be able to do, *e.g.*, `python setup.py test -m desiutil.test.test_bitmask` in order to test individual modules.  There is simple, generic boilerplate that can be added to any test module to enable this.

weaverba137:
Tests are passing, so I'm going to merge momentarily.

PR #51, opened 2016-11-23
Skyplot

fjaviersanchez:
Added utility to plot sky maps. It requires Basemap (which is installable via pip install). I also included a notebook to show the usage.


fjaviersanchez:
Updated requirements

PR #52, opened 2016-11-23
Skyplot

fjaviersanchez:
Added utility to plot sky map. It requires Basemap. Notebook also included in doc/nb.


sbailey:
This could be quite useful for QA plots, but it comes with new dependencies: in addition to `basemap`, it also requires `healpy` if using `pixshape='healpix'`.  These do appear to be sufficiently isolated that the rest of desiutil can be used without them, and they are only needed if you are actually using this function (and for healpy, only if you are using `pixshape='healpix'`).

Travis tests are failing with a segfault, which is surprisingly extreme.  Is that possibly a side effect of Travis not having an X11 environment by default?

Can you make `ra` and `dec` optionally plain numpy vectors with an implicit unit of degrees?  Other parameters `center_longitude`, `epsi`, and `radius` are already plain float values with implicit units of degrees.  If the user provides units, great, use them.  But it would be convenient to be able to pass in normal arrays and have that default to degrees.

Please format the `plot_sky` docstring to have a max of 80 characters.  That helps old-school folks like me who still use the command line on a 80 character wide window, and even makes it more readable to jupyter notebooks too.  :)

`nside` option in docstring is indented too far and I almost missed it.

When giving both `data` and `pix_shape='healpix'`, is the plotted result the sum of the weights (data) or the mean of the data?  Or something else?

fjaviersanchez:
Thanks for the suggestions! I'll update that. I don't know what the problem with Travis is... In principle it shouldn't really matter that it doesn't have a X11 since in the tests the `matplotlib('agg')` option is activated. I am testing the plots without `plot_sky` to see if it is something related to `basemap` itself. When you choose `'healpix'` and `data!=None` the plotted result is in units of counts per square-degree.

fjaviersanchez:
Locally the tests run perfectly in both `python-3.5` and `python-2.7` by doing `python setup.py test`. However, the `plot_sky` tests causes a segfault in Travis-CI so I commented them out. It is possible to uncomment them and proceed with the tests if the user wants to do so.

sbailey:
@weaverba137 is the gatekeeper for desiutil so let's wait for him to get back from the break to review and comment before merging.

You could use decorators like
```
@unittest.skipIf('TRAVIS_JOB_ID' in os.environ, 'Skipping test on Travis')
def test_blatfoo(self):
    ...
```
to skip the tests on Travis while still running them locally without having to edit the file each time.  

Please add something like that also to skip the tests if `healpy` or `basemap` aren't installed, e.g.
```
try:
    import mpl_toolkits.basemap
    have_basemap = True
except ImportError:
    have_basemap = False

...
@unittest.skipIf(have_basemap == False, 'Skipping tests of plot_sky that require basemap to be installed first')
def test_blatfoo(self):
    ...
```

I think you can chain decorators to separately skip based upon healpy, basemap, and Travis, but I haven't tried that.

fjaviersanchez:
Since the tests don't require `healpy` I didn't include it in `skiptest` but I included a clause to skip if the environment is `Travis` or if `basemap` is not installed.

weaverba137:
I don't understand why this has to be in desiutil.

dkirkby:
That was Stephen's suggestion.  I was originally thinking of desimodel (since that's where the tiles live) or desisurvey.  I guess the question is whether desiutil is primary used as an import or also provides high-level scripts of general use.

sbailey:
I was thinking that we might use it from multiple packages: desitarget for plotting target densities, desimodel for plotting tile information, desisurvey for plotting observing progress or observing conditions per tile, desispec for plotting QA metrics per brick, etc.  That would drive the code to our lowest level dependency of utility code, i.e. desiutil, which already has a plotting utility module.

geordie666:
I'd second this from the perspective of desitarget. I think a "go-to" code for plotting sky density maps is generally useful.

weaverba137:
OK, fair enough, but I'm not particularly happy about the new dependencies that this brings in.  Who is maintaining the complete list of dependencies these days?

weaverba137:
Without even looking at the code, I can already see some problems:

1. doc/changes.rst has not been updated.
2. `python setup.py build_sphinx` is throwing errors.
3. I don't understand why tests are being skipped because they *conflict* with Travis.  What, *exactly* does that mean?  Wouldn't it be better to resolve the conflicts?

py/desiutil/plots.py
```diff
@@ -118,3 +118,251 @@ def plot_slices(x, y, x_lo, x_hi, y_cut, num_slices=5, min_count=100, axis=None,
...
+    -------
+   :class:`matplotlib.axes.Axes` 
+       The Axes object for the plot. It creates a figure if
+       there was no previous figure and if data is not provided
+       it returns counts per square-degree if healpix
+       or square pixels are created. If you choose the option ellipse
+       it plots as many ellipses as ra,dec points provided (it may be slow).
+    """
+    import matplotlib
+    if test_travis:
```
weaverba137: If you're skipping tests because of "conflicts" with Travis, why do you have this?

py/desiutil/plots.py
```diff
@@ -118,3 +118,251 @@ def plot_slices(x, y, x_lo, x_hi, y_cut, num_slices=5, min_count=100, axis=None,
...
+
+        # Set axes limits to fit map region.
+        self.set_axes_limits(ax=ax)
+
+        return poly
+
+    if pix_shape not in ['ellipse','healpix','square']:
+        print('Pixel shape invalid, try ellipse, healpix or square')
+    if discrete_colors:
+        if(data is None):
```
weaverba137: Why do you need parentheses here?

py/desiutil/plots.py
```diff
@@ -118,3 +118,251 @@ def plot_slices(x, y, x_lo, x_hi, y_cut, num_slices=5, min_count=100, axis=None,
...
+            print('Error discrete_colors expects data!=None')
```
weaverba137: Why `print()`? Shouldn't this raise an exception?

py/desiutil/plots.py
```diff
@@ -118,3 +118,251 @@ def plot_slices(x, y, x_lo, x_hi, y_cut, num_slices=5, min_count=100, axis=None,
...
+            cmap = plt.get_cmap(cmap)
+            cmaplist = [cmap(i) for i in range(cmap.N)]
+            cmap = cmap.from_list('Custom cmap', cmaplist, cmap.N)
+            # define the bins and normalize
+            bounds = np.linspace(min_color,max_color,nsteps)
+            norm = matplotlib.colors.BoundaryNorm(bounds, cmap.N)
+    else:
+        cmap = plt.get_cmap(cmap)
+        norm = None
+    if(pix_shape=='healpix'):
```
weaverba137: Why do you need parentheses here?

py/desiutil/plots.py
```diff
@@ -118,3 +118,251 @@ def plot_slices(x, y, x_lo, x_hi, y_cut, num_slices=5, min_count=100, axis=None,
...
+        import healpy as hp
+        # get pixel area in degrees
+        pixel_area = hp.pixelfunc.nside2pixarea(nside, degrees=True)
+        #avoid pixels which may cause polygons to wrap around workaround
+        drawing_mask = np.logical_and(np.fabs(ra-180-center_longitude)>2*np.sqrt(pixel_area)+epsi,np.fabs(ra+180-center_longitude)>2*np.sqrt(pixel_area)+epsi)
```
weaverba137: Put this on multiple lines, it is too hard to read.

py/desiutil/plots.py
```diff
@@ -118,3 +118,251 @@ def plot_slices(x, y, x_lo, x_hi, y_cut, num_slices=5, min_count=100, axis=None,
...
+        if not hide_galactic_plane:
+            # generate vector in galactic coordinates and convert to equatorial coordinates
+            galactic_l = np.linspace(0, 2*np.pi, 1000)
+            galactic_plane = SkyCoord(l=galactic_l*u.radian, b=np.zeros_like(galactic_l)*u.radian, frame='galactic').fk5
+            # project to map coordinates
+            galactic_x, galactic_y = m(galactic_plane.ra.degree, galactic_plane.dec.degree)
+            m.scatter(galactic_x, galactic_y, marker='.', s=2, c='k')
+        # Add a colorbar for the PolyCollection
+        plt.colorbar(coll, orientation='horizontal',cmap=cmap, norm=norm, spacing='proportional', pad=0.01, aspect=40, label=label)
+    if(pix_shape=='square'):
```
weaverba137: Why do you need parentheses here?

py/desiutil/test/test_plots.py
```diff
@@ -29,11 +30,13 @@ class TestPlots(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         cls.plot_file = 'test.png'
-
+    #    cls.plot_file2 = 'test_sky.png'
```
weaverba137: Why are you adding this only to comment it out?

requirements.txt
```diff
@@ -1,5 +1,6 @@
 setuptools
 requests
 pyyaml
-matplotlib
```
weaverba137: Are you QUITE CERTAIN that matplotlib is no longer required?  Please demonstrate how this PR eliminates all possible references to matplotlib.

py/desiutil/test/test_plots.py
```diff
@@ -8,13 +8,14 @@
...
 import numpy as np
-
+import sys
 # Set non-interactive backend for Travis
 import matplotlib
 matplotlib.use('agg')
 import matplotlib.pyplot as plt
 
 from ..plots import plot_slices
+#from ..plots import plot_sky
```
weaverba137: You've commented out this import, but that means that `test_plot_sky()` will be broken on systems that are *not* Travis.

py/desiutil/plots.py
weaverba137: If you look in the .travis.yml file for desispec, there's a (commented out) example of how to start an X11 instance on Travis.  However, the 'agg' environment should be functional outside of X11.  Are you certain that the 'agg' environment works even when you're *not* on Travis?

fjaviersanchez: I tested the 'agg' environment and it worked in my laptop and in a local server but, it kept crashing in Travis. I am testing the fake X11 on Travis now, thanks for the tip!

requirements.txt
fjaviersanchez: Yes, you are totally right. I was confused because `basemap` requires `matplotlib` so if you have one you should have the other but, if the user doesn't want to use `plot_sky` then, `basemap` is not required and therefore `matplotlib` is no granted. I have changed this.

sbailey:
At the same time, we may end up wanting to use this code for automated QA plots generated by the pipeline within a batch environment without X11.  i.e. in this case Travis lacking X11 could be considered a feature not a bug and we should make sure that the code can be used to generate and save a plot even without an X11 connection.  Enabling the fake X11 server within Travis should only be a temporary workaround on the way to allowing this code to not require X11.

fjaviersanchez:
The problem in Travis happens when I want to call `plt.savefig` the second time and this causes the segmentation fault. Any thoughts?

dkirkby:
Where are you telling savefig() to write the two files?

weaverba137:
I usually call `figure.savefig()`, rather than `pyplot.savefig()`.

dkirkby:
See [this example](https://gist.github.com/odyniec/d4ea0959d4e0ba17a980) of using tempfile for unit tests that produce output (although I don't see why this would only fail on the second write).

fjaviersanchez:
@dkirkby @weaverba137 I tried both options but none of them worked. I just omitted the writing part if the working environments is Travis. I still don't know why it doesn't have any problems with the first `savefig` but when it tries to save for the second time it crashes.

weaverba137:
Please add a sentence describing this change to doc/changes.rst before merging.

weaverba137:
I'm sorry, but this merge is *not* the time to set up for a new tag.  We are not necessarily releasing today.

doc/changes.rst
```diff
@@ -7,7 +7,6 @@ Change Log
 
 *
 
-
```
weaverba137: I don't see how removing one line addresses my concern about not trying to set up for a tag in a merge.

fjaviersanchez: Sorry, my bad. I didn't see your comment! Where do I have to write the updates?

fjaviersanchez: Alright, I think I changed it in the right place. Please, let me know if I have to change it again. Thanks.

weaverba137: I'm sorry, but now it looks like you deleted all the notes that were already under 1.9.3.  I strongly suggest you revert to the master version, then add in your *one line* about the change under 1.9.3 (unreleased).

fjaviersanchez:
The notes under 1.9.3 were mine. I replaced them with two lines. I will change it to one line.

.travis.yml
```diff
@@ -69,7 +69,7 @@ matrix:
 #           env: MAIN_CMD='pep8' SETUP_CMD='--count py/desiutil'
 
 before_install:
-    - curl ipinfo.io
+    - curl ipinfo.io
```
weaverba137: It now looks like you're only changing .travis.yml by adding a whitespace to the end of a line.  Please revert .travis.yml if this is really the only change.

py/desiutil/plots.py
```diff
@@ -118,3 +119,248 @@ def plot_slices(x, y, x_lo, x_hi, y_cut, num_slices=5, min_count=100, axis=None,
...
+    from matplotlib.patches import Polygon
+    from mpl_toolkits.basemap import pyproj
+    from mpl_toolkits.basemap import Basemap
+    
+    #---------
+    # Add ellipses to Basemap
+    #--------
+
+    class Basemap(Basemap):
+        #Code from http://stackoverflow.com/questions/8161144/drawing-ellipses-on-matplotlib-basemap-projections
```
weaverba137: You should add class documentation here, not a comment.  Are you *really* sure you want to also call the subclass 'Basemap'?

fjaviersanchez:
I added some documentation about the `Basemap` class. I tried to use the ellipses from `matplotlib` but they don't project well into the map, that's why I had to extend `Basemap`. The function `ellipses` inside the `Basemap` class extend the original `Basemap` from `mpl_toolkits.basemap` to draw ellipses but since I didn't want to add dependencies in the package for users that won't use `plot_sky` I couldn't think about any other way of doing this besides embedding the class extension in `plot_sky`.

weaverba137:
I think this is ready.  Does anyone else have any comments?

dkirkby:
I am taking a look at it now...

dkirkby:
To browse the new notebook before the merge visit https://github.com/fjaviersanchez/desiutil/blob/skyplot/doc/nb/SkyMapExamples.ipynb

weaverba137:
OK, here's the first review item: let's make sure Travis tests are actually passing.

weaverba137:
Re: requirements.txt.  This file is how dependencies are installed in Travis tests (see the .travis.yml file).  So if the absence of a package is causing Travis tests to fail, it should be added to requirements.txt.

dkirkby:
Travis tests should be passing now.  I included some runnable tests in the docstring to `desiutil.plots.prepare_data()` that are not currently being run by `python setup.py test` but I have checked them using:
```
pytest --doctest-modules py/desiutil/plots.py
```

weaverba137:
Good, I'll review this momentarily.

.gitignore
```diff
@@ -5,6 +5,7 @@ __pycache__
 # Other generated files
 htmlcov
 .coverage
+py/desiutil/.cache
```
weaverba137: How is this file or directory generated?  Remember that Python packages tend to get installed read-only, so this is not necessarily good practice.

doc/changes.rst
```diff
@@ -5,7 +5,9 @@ Change Log
 1.9.3 (unreleased)
 ------------------
 
-* No changes yet.
+* Added :func:`desiutil.plots.plot_sky` to make all-sky plots (`PR #52`_).
```
weaverba137: It looks like a lot more work was done than what this sentence implies.  Also, in all the rewriting, does that function still exit, or apply?

.gitignore
dkirkby: This file was generated by running `pytest --doctest-modules py/desiutil/plots.py`, which isn't normally invoked by travis or `python setup.py test` (but perhaps should be?)  In any case, I just checked that `pytest` runs fine when `py/desiutil/` is read only.

doc/changes.rst
dkirkby: Thanks for catching this.  I just updated the change log.

weaverba137:
Unless there's anything anyone else wants to add, I think this is ready to merge.

dkirkby:
@sbailey @fjaviersanchez ok with you to merge now?

sbailey:
Come on, we're almost up to 100 github generated emails for this thread, surely someone else has something to say to get us over that threshold!  Just kidding.  Looks good, merging now.

(I may have future requests about RA increasing left vs. right and negative RA axis labels, but those don't need to be a blocking factor for getting this in now).

dkirkby:
Is it normal that this PR was merged as a single commit back to master?  It looks like we lost a lot of detailed commit history.

dkirkby:
I guess @sbailey chose the [squash and merge option](https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits)?  In any case, all of my commit messages are gone (as is any record that I worked on this).

sbailey:
Yes, I used squash and merge (which was the default option).  I think all of the individual log messages are preserved in the mega-merge log message a918c15da but I didn't realize that it lost track of who had contributed what (apologies).  I dislike git's ability to rewrite history so easily...

dkirkby:
Clicking on "Merge Pull Request" is supposed to create a merge commit by default and should never rewrite history.  Other merge options can normally only be selected via this pull-down menu, which does a reasonable job of explaining the consequences:

merge

sbailey:
Empirically I have seen that sometimes "squash and merge" is the default.  It seems to come and go and I haven't figured out why (I thought it might be due to the complexity of the PR, but haven't been able to reproduce that).  I was surprised when it first showed up as a default option a few months ago but figured that it was the new recommended workflow for github and went with it.  From this conversation I'm inclined to go back to always doing the merge commit unless
1. a single developer contributed to the PR (otherwise the loss of contribution history is particularly irritating)
2. the PR has a clear simple new feature or bug fix (and not N>>1 features in M>>1 commits that are all finally contributed via a single PR; that's bad practice but it happens anyway)
3. most of the commits were minor fixes or doc updates that just clutter the overall history

Apologies for the loss of github activity credit.

weaverba137:
I would also note that this was a merge between two different forks, not two branches in the same fork, so that may have affected the default merge pattern.

dkirkby:
A default "squash & merge" sounds like a github bug, based on [this](https://help.github.com/articles/about-pull-request-merges/):
```
When you click the default Merge pull request option on your pull request on GitHub,
all your commits from the feature branch are added to the base branch in a merge commit.
```
You can try a merge between forks yourself by re-opening #51 and clicking on the merge pull-down (but don't actually re-do the merge).

If we care about this, we can disable the "squash & merge" option in the settings for each repo.

sbailey:
desihub/desisim#199 is an example of a PR that has defaulted to a Squash and merge:

screen shot 2016-12-22 at 1 22 10 pm

But desihub/desisurvey#14 defaulted to a normal merge commit:

screen shot 2016-12-22 at 1 25 34 pm

PR #54, opened 2017-02-14
add new sklearn.GaussianMixtureModel functionality

moustakas:
This PR adds a new module semi-arbitrarily called ```sklearn``` which contains the ```GaussianMixtureModel``` class taken from sklearn.  This class was previously in ```desisim.templates``` but now it is needed by both the template-generating code *and* ```desitarget.mock.io``` (as part of https://github.com/desihub/desitarget/issues/136; see also https://github.com/desihub/desitarget/issues/106 and [dr2_mag_distributions.ipynb](https://github.com/desihub/desitarget/blob/master/doc/dr2_mag_distributions.ipynb).

I called it sklearn in case we decide to pull in other functionality without introducing an unnecessary dependence but we can easily change the name.

[Thanks to @dkirkby, who originally pulled this piece out of sklearn.]

### DIFF
--- a/py/desiutil/sklearn.py
+++ b/py/desiutil/sklearn.py
@@ -1,0 +1,72 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+"""
+================
+desiutil.sklearn
+================
+
+Useful functions from the sklearn python package.
+"""
+from __future__ import (print_function, absolute_import, division,
+                        unicode_literals)
+
+class GaussianMixtureModel(object):
+    """Read and sample from a pre-defined Gaussian mixture model.
+
+    """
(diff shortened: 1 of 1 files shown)


weaverba137:
@moustakas, You requested my review on this, but didn't actually give me a chance to review it.  For urgent cases like this, please ping me if I don't see it right away.

Could you *please* update the change log, `doc/changes.rst`, and add desiutil.sklearn to the API documentation, `doc/api.rst`?  Please also verify that the documentation builds without errors.

weaverba137:
PS, the documentation of the module itself is extremely sparse.

weaverba137:
@moustakas are you listening?

geordie666:
@moustakas <https://github.com/moustakas> is traveling today, so he's only
in intermittent contact...

PR #56, opened 2017-02-28
add log module to desiutil

weaverba137:
This PR adds the logging module to desiutil so that it can be used by other DESI packages (besides desispec) without creating circular imports.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,10 +5,14 @@ Change Log
 1.9.3 (unreleased)
 ------------------
 
-* Added new :mod:`desiutil.sklearn` module and :class:`distutils.sklearn.GaussianMixtureModel` class to save and sample from a Gaussian mixture model.
-* Added new functions for creating all-sky maps (`PR #52`_) with an accompanying tutorial notebook in `doc/nb/`.
+* Added new :mod:`desiutil.sklearn` module and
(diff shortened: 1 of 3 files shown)


sbailey:
Question: our log messages use colon (`:`) delimiters.  The new `timestamp` option here adds a timestamp that includes colons as non-delimiters.  Do we care?  Options:
  1. We don't care.  I'm not aware of anyone actually using the colons as delimiters to parse log messages, and even with the timestamps we could use something like `levelname, filename, lineno, funcname = message.split(':', 4)[0:4]` that would work either way
  2. Change to using something other than colons as delimiters
  3. Change the timestamp to use something slightly non-standard but still readable like 'datefmt='%Y-%m-%dT%H.%M.%S' (. instead of : in the time)

I don't particularly have a preference but I thought I'd raise the question.  OK to merge as-is if you prefer option (1)

moustakas:
After this is merged could we please tag `desiutil`?  Both `desisim` and `desitarget` need the new `sklearn` module, which is currently only in the master branch.

weaverba137:
@sbailey, there's a fourth option: add an optional delimiter.  Most people won't use it, & won't care, since timestamps are not default.

@moustakas, yes, that's a good idea, but I'm pretty busy at a conference, so deploying the tag could take a while.

PR #61, opened 2017-04-04
Minimal requirements

### LINKED ISSUE #60: Please ensure that basemap is installed correctly during Travis tests, and add tests for plots.py
The module `desiutil.plots` needs [basemap](https://github.com/matplotlib/basemap) for some functions.  However, the [basemap documentation](http://matplotlib.org/basemap/users/installing.html) suggests that the install procedure is somewhat more complicated than just `pip install git+https://github.com/matplotlib/basemap.git`.

Please ensure that:

* basemap is being installed correctly during Travis tests.
* That there are at least some tests to verify that basemap is installed correctly.
* Finally, please provide instructions on how to install basemap on NERSC systems, and ensure that it is being installed correctly in the DESI+Anaconda system.

weaverba137:
This PR:

* Closes #60 
* Adds documentation to distinguish between required and optional dependencies.
* Improves documentation of plots.py module.
* Improves test coverage of plots.py module.
* Temporarily removes an experimental macOS test.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -18,7 +18,8 @@ Required Dependencies
 These packages must be installed for desiutil to work properly:
 
 * `pyyaml <http://pyyaml.org/>`_
-* `astropy <http://www.astropy.org/`_
+* `astropy <http://www.astropy.org/>`_
+
   - This implies a dependency on `NumPy <http://www.numpy.org>`_
 
 Optional Dependencies
@@ -30,6 +31,9 @@ need:
 * `matplotlib <http://matplotlib.org/>`_
 * `healpy <https://healpy.readthedocs.io/en/latest/>`_
 * `basemap <http://matplotlib.org/basemap/>`_
+
+  - Note that basemap's installation is non-trivial.  See the documentation
+    linked to above.
 
 Contents
 ========
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -5,7 +5,31 @@ Welcome to desiutil's documentation!
 Introduction
 ============
 
-This is the documentation for desiutil.
+desiutil is a set of low-level utilities used by all DESI_ packages.
+
+.. _DESI: https://desi.lbl.gov
+
+Dependencies
+============
+
+Required Dependencies
++++++++++++++++++++++
+
+These packages must be installed for desiutil to work properly:
+
+* `pyyaml <http://pyyaml.org/>`_
+* `astropy <http://www.astropy.org/`_
+  - This implies a dependency on `NumPy <http://www.numpy.org>`_
+
+Optional Dependencies
++++++++++++++++++++++
+
+If you want to use the plotting utilities in :mod:`desiutil.plots`, you will
+need:
+
+* `matplotlib <http://matplotlib.org/>`_
+* `healpy <https://healpy.readthedocs.io/en/latest/>`_
+* `basemap <http://matplotlib.org/basemap/>`_
 
 Contents
 ========
--- a/py/desiutil/plots.py
+++ b/py/desiutil/plots.py
@@ -130,8 +130,6 @@ class MaskedArrayWithLimits(numpy.ma.MaskedArray):
     This class accepts the same arguments as
     :class:`~numpy.ma.MaskedArray`.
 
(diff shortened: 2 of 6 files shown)


doc/index.rst
```diff
@@ -5,7 +5,35 @@ Welcome to desiutil's documentation!
...
+  - This implies a dependency on `NumPy <http://www.numpy.org>`_
+
+Optional Dependencies
++++++++++++++++++++++
+
+If you want to use the plotting utilities in :mod:`desiutil.plots`, you will
+need:
+
+* `matplotlib <http://matplotlib.org/>`_
+* `healpy <https://healpy.readthedocs.io/en/latest/>`_
```
dkirkby: healpy is only required for one type of plot, so its doubly optional.

py/desiutil/plots.py
```diff
@@ -125,26 +125,60 @@ def plot_slices(x, y, x_lo, x_hi, y_cut, num_slices=5, min_count=100, axis=None,
...
+        # else:
+        #     obj.vmax = None
         return obj
 
-    def __array_finalize__(self, obj):
-        if obj is None: return
-        numpy.ma.MaskedArray.__array_finalize__(self, obj)
-        self.vmin = getattr(obj, 'vmin', None)
-        self.vmax = getattr(obj, 'vmax', None)
+    # def __array_finalize__(self, obj):
```
dkirkby: I guess you checked this dunder method isn't actually needed? (I just copied the example)

py/desiutil/plots.py
```diff
@@ -206,55 +272,32 @@ def prepare_data(data, mask=None, clip_lo=None, clip_hi=None,
...
         except AttributeError:
             # Nothing is masked by default.
-            mask = np.zeros_like(data, dtype=bool)
+            cmask = np.zeros_like(data, dtype=bool)
     else:
-        mask = np.asarray(mask)
-        if mask.shape != data.shape:
+        #
+        # Make every effort to ensure that modifying the mask of the output
+        # does not modify the input mask.
```
dkirkby: How does renaming mask to cmask accomplish this?

doc/index.rst
weaverba137: I *think* it might be two plots, or at least healpy appears in two functions.

py/desiutil/plots.py
weaverba137: I'm not sure what a "dunder" method is, but I do know that the original code was *not* working as expected, and the new code is.  Please see the unit tests.

py/desiutil/plots.py
weaverba137: It's not a renaming, it's a copy.  That function was modifying its inputs in unexpected ways.  In the original code, if you input a masked array, and then input the *same* masked array into the function *again*, the inputs will not be the same.

sbailey: Note: `np.asarray(x)` will not make a copy if `x` is already an array, while `np.array(x)` will make a copy.  Both will promote lists to arrays if needed (thus giving more input flexibility than `x.copy()`).  i.e. I think the code below could have been implemented as simply `cmask = np.array(mask)` without needing try/except.

doc/index.rst
weaverba137: Indeed, healpy is required by *both* `plot_healpix_map()` and `plot_sky_binned()`.

py/desiutil/plots.py
weaverba137: Indeed, the simplified `cmask = np.array(mask)` appears to work, and is simpler.

weaverba137:
Will merge as soon as the last round of tests pass.

doc/index.rst
dkirkby: Yes, you're right.  But healpix-deniers can still make pretty plots without installing healpy if they insist.

py/desiutil/plots.py
dkirkby: Ok, thanks.  ([dunder](http://www.python-course.eu/python3_magic_methods.php) ~ "double underscore").

py/desiutil/plots.py
dkirkby: Great!

PR #65, opened 2017-04-13
fix desiInstall on cori

sbailey:
This PR fixes desiInstall for cori, but i'm not sure it is the best solution (see below)

Previously desiInstall looked for `modulecmd` in `/opt/modules/${MODULE_VERSION}/bin/modulecmd`.  That is true for some but not all versions of modules at NERSC; in particular it is not true for the current version of modules on cori (3.2.10.5).  This PR works around that by also checking `/opt/cray/pe/modules/${MODULE_VERSION}/bin/modulecmd` instead.

For consideration: I think it would be better for this whole section to use `distutils.spawn.find_executable('modulecmd')` rather than special casing the likely location on multiple machines.  So I'm offering the minimal fix in this PR, but perhaps we should use `find_executable` or similar instead.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -8,6 +8,7 @@ Change Log
 * Get .travis.yml file and other components ready for Python 3.6.
 * Increase test coverage in a few areas.
 * Make basemap_ an optional dependency (PR `#61`_).
+* fix desiInstall on cori
 
 .. _basemap: http://matplotlib.org/basemap/
 .. _`#61`: https://github.com/desihub/desiutil/pull/61
--- a/py/desiutil/modules.py
+++ b/py/desiutil/modules.py
@@ -72,7 +72,12 @@ def init_modules(moduleshome=None, method=False, command=False):
         #
         # This is probably one of the primary NERSC systems, edison or cori.
         #
-        modulecmd = ['/opt/modules/{MODULE_VERSION}/bin/modulecmd'.format(**os.environ), 'python']
+        tmpcmd = '/opt/modules/{MODULE_VERSION}/bin/modulecmd'.format(**os.environ) 
+        # if that doesn't exist, try in /opt/cray/pe/modules (cori)
+        if not os.path.exists(tmpcmd):
+            tmpcmd = '/opt/cray/pe/modules/{MODULE_VERSION}/bin/modulecmd'.format(**os.environ) 
+
+        modulecmd = [tmpcmd, 'python']
         os.environ['MODULE_VERSION_STACK'] = os.environ['MODULE_VERSION']
     elif os.path.exists(os.path.join(moduleshome, 'modulecmd.tcl')):
         #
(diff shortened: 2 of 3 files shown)


weaverba137:
I would consider using [shutil.which](https://docs.python.org/dev/library/shutil.html#shutil.which), although that is only available in Python 3.  distutils is kind of ancient, and its API documentation isn't great.

Also, the unit test should be fixed.

sbailey:
Switching to `shutil.which` is fine with me.  We effectively only use desiInstall at NERSC and only use python 3.5 there, so its fine if desiInstall doesn't support py2.7.

Do you have suggestions for fixing the unit tests?  It is currently a rather delicate dance of environment variables and config dirs, and after switching to `shutil.which` then `init_modules` could find `modulecmd` or `modulecmd.tcl` at an arbitrary location on a system or perhaps not at all.

I'm ok with a simpler pragmatic test:
  * if $NERSC_HOST is set then the test should be running at NERSC and `init_modules` should be able to find `modulecmd`
  * otherwise it should either find `modulecmd`, `modulecmd.tcl`, or raise an exception.

weaverba137:
Wasn't Ted talking on Tuesday about a situation where Python 2.7 was required for something at NERSC?  If that's something we'd never actually need desiInstall for, then that's fine, but I wanted to make sure we had covered that case.

sbailey:
He's setting up a py27 desiconda stack for the imaging surveys dependencies, but they don't use desiInstall.  He also had do some py27 vs. py35 craziness for galsim since we need it under 3.5 but it uses the scons config/install package that only works under 2.7.  But at the end of the day, the only place where we actually run desiInstall is 3.5.  Or so I think...

weaverba137:
OK, I can take this over if you want.

sbailey:
> OK, I can take this over if you want.

That would be appreciated, thanks.

On second thought about `distutils.spawn.find_executable` vs. `shutil.which`: even though distutils is crufty, it is part of the core language distribution (not an abandoned external package) and that call does what we want for both py2.7 and 3.5.  Let's use that in order to not arbitrarily break py2.7, while leaving a comment in the code that if for whatever reason this stops working, trying using `shutils.which` instead.  Or if you hit any trouble using `distutils.spawn.find_executable` now, go ahead and switch to `shutil.which` if that works better.  i.e. let's not break desiInstall on py2.7 out of an aesthetic preference for one single-line vs. another single-line call.  Who knows? the imaging surveys may want some sliver of our code...

weaverba137:
I think this might be the most elegant solution:

```python
try:
    from shutil import which
except ImportError:
    # shutil.which is Python 3
    from distutil.spawn import find_executable as which
```

As far as I can tell the two functions have an identical API, at least for our purposes.

weaverba137:
Tests have passed.  @sbailey, last call for comments before I merge.

sbailey:
go for it

Issue #82, opened 2017-09-07
What is happening with the durhamplots branch?


weaverba137:
Are you planning do to any further work on this branch?  It appears that you made one trivial change, though you should document the fact that the function now returns more than one value.

forero:
I believe I won't add new changes (now that I've left Durham). I will document and start a PR.

weaverba137:
Can we get this done?  Should take 10 minutes of work at most.

forero:
Done. I am just waiting for @dkirkby to review that tiny change and be sure that I am not breaking something important.

weaverba137:
Excellent, thank you.

weaverba137:
Closed by #85.

PR #84, opened 2017-09-07
Fix bugs that caused bricks to be larger than bricksize.

dstndstn:
Fixes some implicit assumptions that bricksize evenly divides 180 degrees:
* don't assume the last brick is at the north pole (Dec=90)
* avoid Dec edges exceeding 90 degrees
* avoid Dec array indexing error
* fix handling of the equator in computing required number of bricks in a row

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -9,9 +9,13 @@ Change Log
   in an installed package, as opposed to a git checkout (PR `#80`_).
 * Fixed bug in :meth:`desiutil.brick.Bricks.brick_radec` handling scalar inputs
   (PR `#81`_).
+* Fixed bugs that could cause bricks to be slightly too big, and that
+  incorrectly special-cased the north pole with brick sizes that don't
+  evenly divide 180 degrees (PR `#82`_).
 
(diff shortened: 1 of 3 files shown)


doc/changes.rst
```diff
@@ -9,9 +9,13 @@ Change Log
   in an installed package, as opposed to a git checkout (PR `#80`_).
 * Fixed bug in :meth:`desiutil.brick.Bricks.brick_radec` handling scalar inputs
   (PR `#81`_).
+* Fixed bugs that could cause bricks to be slightly too big, and that
```
weaverba137: This PR is #84, not #82.

py/desiutil/test/test_brick.py
```diff
@@ -85,6 +85,43 @@ def test_brickvertices_wrap(self):
         self.assertEqual(np.max(b1[:,0])-np.min(b1[:,0]), 360.)
         self.assertTrue(np.all(b1[:,1] >= -90.))
 
+    def test_uneven_bricksize(self):
```
weaverba137: Please add a `__doc__` string.

weaverba137:
One non-critical test is stalled, merging anyway.

Issue #68, opened 2017-04-26
requested enhancements to plots.py


moustakas:
The `desiutil.plots` module is fantastic but for QA it would be helpful if the following enhancements could be implemented:
* In `plot_sky_binned` (and possibly elsewhere) could `grid_data` be returned to the user?  For example, in addition to the skymap I'd like to separately look at the (pixelized) target densities (e.g., to make a histogram of the values stored in `grid_data`).  I could of course call the snippet of healpy code myself but that would be duplicating code.

* More challengingly is the question of how to deal with edges in the skymap (and the resulting target densities).  For example, if the input catalog doesn't fully span an edge pixel then the target density will be underestimated since it's being divided by the same area as the fully populated pixels.  The solution to this may be non-trivial and outside the scope of this module but nevertheless it would be great to solve this so that the QA is as useful as possible.

moustakas:
I believe that @geordie666 is tackling the second requested feature as part of his targeting QA effort (https://github.com/desihub/desitarget/issues/21).  Along these lines, this notebook by @dkirkby should be very useful, as it shows how to generate a set of random points given a set of healpix pixels -- 
  https://nbviewer.jupyter.org/gist/anonymous/3625c69cd642d37cc9b3c6f4aaf86749

My thinking was to do the following for a wide range (all?) of nside values:
1. Find all the healpixels in the DESI footprint using `desimodel.footprint.tiles2pix`.
2. Throw down a large number of random points in these healpixels using @dkirkby's method.
3. For each healpixel, use `desimodel.footprint.is_point_in_desi` to find the fraction of random points in our footprint.  Store this weight in a light-weight FITS table which will be included in `desiutil`.
4. Read, cache, and use these weights by default when calling `plot_sky_binned`.

@geordie666 may have a different approach but this was my thinking.

geordie666:
Yeah, I started looking at this and I'll finish it this week. For all the pixels in the (~14k sq. deg.) DESI footprint, @dkirkby's approach isn't much quicker than populating the entire sphere at relevant declinations. But, I think I can populate just the fractional pixels (i.e. the pixels that are partially in and partially out of the DESI footprint). Then things are much faster.

geordie666:
Oh, I should add that we don't need to do this for a wide range of `NSIDE` values, just some high `NSIDE` (say `NSIDE` == 256). 

[...]

PR #85, opened 2017-09-27
plot map returns grid data besides basemap

forero:
This minor change allows `grid_data` manipulation to make and compare histograms of different maps.


weaverba137:
@geordie666, please verify that this fixes #68.

geordie666:
I believe that this will fix #68 when merged, yes. However, I'd recommend at least the following under review:

1. The documentation (https://github.com/desihub/desiutil/blob/15b74974621cdfdd3531557e539b616872951150/py/desiutil/plots.py#L885-L889) needs changed to include `grid_data` as a return

2. As the addition of an extra returned array will potentially break a lot of downstream code, I'd be tempted to add an input `return_grid_data=False` that will only return the `grid_data` if prompted, e.g.:

```
if return_grid_data is True:
    return basemap, grid_data

return basemap
```
That's just personal taste, though. People could also just update their calls to `plot_sky_binned` as and when they break.

weaverba137:
Please add a brief note about this to the `changes.rst` file before merging.

moustakas:
I agree with @geordie666 that returning `grid_data` should be optional, not the default, although I *can* deal with code breakage if that's inconvenient.

sbailey:
I'd like to make new tags today, so I made the changes suggested by @geordie666 (make new grid_data return optional) and @weaverba137 (update changes.rst) and will merge after Travis tests pass again.

doc/changes.rst
```diff
@@ -13,11 +13,14 @@ Change Log
   incorrectly special-cased the north pole with brick sizes that don't
   evenly divide 180 degrees (PR `#84`_).
 * Added tests of :mod:`desiutil.sklearn` (PR `#86`_).
+* Adds ``return_grid_data``` option to ``desiutil.plots.plot_sky_binned()``
```
weaverba137: I think there's a couple of problems here.  This line should be:
```
* Adds ``retrun_grid_data`` option to :func:`desiutil.plots.plot_sky_binned`.
```

dkirkby:
I won't have time to test this today, but it sounds fine to me as long as the default return type has not changed (since I am using the original API in many places).

PR #90, opened 2017-10-17
Update product root

### LINKED ISSUE #88: bootstrapping desiutil & desimodules into new conda environment
We need a way to install a new set of desiutil, desimodules, etc. into a new desiconda environment, e.g.  /global/common/software/desi/cori/desiconda/20170920-1.2.0-spec/ .  The previous instructions at https://desi.lbl.gov/trac/wiki/Pipeline/Install don't work anymore.

A couple issues to keep in mind [this replicates some emails, but I didn't find a ticket with the same info to reference...]

The base directory (/global/common/software/desi/cori/) keeps changing on us.  Each time we think this is the last time, but it probably won't be, so it shouldn't hard coded.  The previous top-level location was /global/common/cori/contrib/desi/ (which NERSC has asked us to migrate away from), but I don't think we ever got that into a fully working example of the directory structure we wanted, so don't take it too literally if it doesn't make sense.

We originally thought that we could decouple desiconda installations from desiutil, desispec, etc. installations, but that didn't work out, requiring that after every new desiconda install (e.g. to upgrade numpy/scipy/astropy/...) we need to reinstall all the desi packages.  i.e. bootstrapping the environment is no longer a rare case to be handled by a single expert; it will be a several-times-a-year thing.

I think we converged on this structure underneath a {basedir}, but I don't think we ever got an example fully fleshed out:
```
desiconda/{version}/
    conda/            - conda packages
    aux/              - compiled libraries releated to this desiconda release
    modulefiles/      - module files for desi packages
        {package}/
            {pkgversion}
startup/
    modulefiles/
        desimodules/  - module files defining sets of desiconda + other packages
            {desimodule_version}
```

Since desimodules picks a desiconda version + a self-consistent set of desispec, desisim, redrock, specsim, etc. packages, its modules file should live *outside* the structure for a single desiconda.  Conversely, since the installation for desiutil/desispec/specsim/etc. is tied to a particular desiconda release, modulefiles for installations tied to that release are contained within that release.

In the end, desi_environment.sh would
```
module use {basedir}/startup/modulefiles
module load desimodules/{version}
```

and `desimodules/{version}` would in turn:
```
module use {basedir}/{desiconda-version}
module load desiconda
module load desiutil/{p.q.r}
module load desispec/{x.y.z}
module load desisim/{a.b.c}
...
```

### LINKED ISSUE #89: raise desiInstall error if desimodel data is missing
`desiInstall desimodel {version}` should fail if there isn't a matching data version in SVN.  Currently it silently installs the code without the data without any warnings or errors or non-zero return codes to indicate that the installed desimodel is problematic.

The fix might be for something on the desimodel side to raise an error that would get noticed by desiInstall.  I haven't looked under the hood.

I discovered this when trying to install desimodel/0.9.0 before remembering to make a matching svn tag.  That tag now exists so you'll need to make a fake code tag without a matching svn tag for reproducing/testing this.

weaverba137:
This PR:

* Fixes #88 
* Fixes #89 

*This branch needs additional tests before merging.*

### DIFF
--- a/LICENSE.rst
+++ b/LICENSE.rst
@@ -1,4 +1,4 @@
-Copyright (c) 2014-2015, DESI Collaboration <desi-data@desi.lbl.gov>
+Copyright (c) 2014-2017, DESI Collaboration <desi-data@desi.lbl.gov>
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without modification,
@@ -9,8 +9,8 @@ are permitted provided that the following conditions are met:
 * Redistributions in binary form must reproduce the above copyright notice, this
   list of conditions and the following disclaimer in the documentation and/or
   other materials provided with the distribution.
-* Neither the name of the Astropy Team nor the names of its contributors may be
-  used to endorse or promote products derived from
+* Neither the name of the DESI Collaboration nor the names of its contributors
+  may be used to endorse or promote products derived from
   this software without specific prior written permission.
 
 THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
--- a/bin/desiBootstrap.sh
+++ b/bin/desiBootstrap.sh
@@ -77,7 +77,7 @@ if [[ -n "${NERSC_HOST}" && -z "${py}" ]]; then
             fi
             py=${dd}/code/desiconda/${anaconda}_conda/bin/python
             if [[ ! -x ${py} ]]; then
-                echo "Python executable not found!"
+                echo "Python executable ${py} not found!"
                 exit 1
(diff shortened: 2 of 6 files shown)


weaverba137:
This PR is ready for review.  I'm satisfied with the tests on /global/common/software, and as a bonus, this supports knl installs (with an added command-line option).

weaverba137:
Please also review https://desi.lbl.gov/trac/wiki/Pipeline/Install, which I have updated.

weaverba137:
I'm going to go ahead & merge.  @sbailey, could you please still review https://desi.lbl.gov/trac/wiki/Pipeline/Install when you get a chance?

PR #92, opened 2017-11-21
Add context-manager functionality to desiutil.log.

### LINKED ISSUE #66: Questions/suggestions for logging module
I have some questions/suggestions about `log.get_logger()` after adapting `desisurvey` and `surveysim` to use it.

Environment variables usually set a default that can be overridden via the API, but `$DESI_LOGLEVEL` overrides the API `level` parameter.  I found this confusing and wonder if we should change this.

A command-line script should be able to set the log-level (e.g., via cmd-line options) and then have all library code respect this level.  I believe the way this is supposed to work is:
```
# script.py
log = get_logger(WARNING)

# module.py
log = get_logger()
log.info('...')
```
I found two issues with how this works in practice:
1. Calling `get_logger()` always sets the log-level to `INFO`, even after a previous call where a different level was specified.
2. Nothing prevents library code from calling `get_logger(DEBUG)` and changing the log level specified by the top-level script.  Library code should probably always call `get_logger()` but we could enforce this by raising an exception on attempts to change an already specified level. Even better, I think, would be to provide separate `get_logger()` (with no level arg) and `set_log_level(level)` methods.

Finally, it could be useful to support a hierarchy of DESI loggers with different levels.  For example, I might want to enable DEBUG messages from the arc calibration only, without being flooded with other DEBUG messages.  In practice, this could be implemented by passing an optional component arg to `get_logger()` that uses dot notation to express hierarchy.  The level of a child component would then default to the lowest level of any parent, and setting a parent level enforces a maximum level for all children.

weaverba137:
This PR closes #66.  Given other stuff going on, this is non-urgent!

Please see the module-level documentation for the `desiutil.log` module for details about how the module currently works.

A few things to note:

* The simplest use case is unchanged.  This is different from what was requested in #66, but at least the existing API is unchanged.  It is not obvious to me how to implement something like `from desiutil.log import log; log.info('blat foo')`.  So rather than spend lots more time working on that, I'd like to get the context manager in.  That should solve a number of problems with the added bonus of not breaking anything already using `desiutil.log`.
* `DESI_LOGLEVEL` still trumps all other methods of setting the log level when `get_logger()` is called.  Even if `get_logger(level)` is used.

Further technical discussion: consider this code:
```python
from desiutil.log import log
log.info('blat foo')
```
How does one *configure* the `log` object without performing a function call?  By hacking the import mechanism?  I honestly don't know.  That's why I'd rather work on the context manager, and just declare this done.


weaverba137:
PS, there are some additional changes under the hood that are not directly related to `desiutil.log`:

* desiInstall now eats its own dogfood by logging with `desiutil.log`, instead of creating its own log object.
* Tests are enhanced with `unittest.mock`, which is now standard in Python 3.

sbailey:
Documenting a conversation with @weaverba137:
> DESI_LOGLEVEL still trumps all other methods of setting the log level when get_logger() is called. Even if get_logger(level) is used.

I think we want it the other way around, that `get_logger(level)` trumps `$DESI_LOGLEVEL`, and that in general code should *not* try to hardcode that override level.  See the Sept 28 comment in #66 for suggested code and rules.

This PR currently still has the unwanted side effect that `get_logger(level)` resets the default level for all future calls to `get_logger()`.  That problem pre-dates this PR, but it is something we want to get away from.  A proposed fix is to generate different logger singletons for each level, e.g. inside `get_logger`:
```python
log = logging.getLogger('desi.{}'.format(level))
# [sent handlers, formatting, etc]
return log
```
@weaverba137 will explore that option more for any gotchas.

weaverba137:
In the latest commits: 

1. If the context manager is called without setting a level, a warning is emitted, but it still works.
2. Calling `get_logger(level)` now trumps `DESI_LOGLEVEL`.
3. String versions of the level names now work, *e.g.* `from desiutil.log import get_logger; log = get_logger('DEBUG')` now works the same as `from desiutil.log import get_logger, DEBUG; log = get_logger(DEBUG)`.
4. Additional error checking is performed on the value of the level.  Previously, only `DESI_LOGLEVEL` values were checked for errors.  If an invalid value of level is detected, a warning is emitted, and the level is set to `INFO`.

I would advocate breaking out the problem of reseting the log level into a separate PR.

weaverba137:
@sbailey, I think this is ready for review & merging.  If you have questions, let's go over them while I'm still in Berkeley.

I actually have a working idea on how to deal the changing log level problem, but that should go in a separate PR.

sbailey:
I don't think this addresses the requests of #66, but it has useful pieces in the right direction so let's go ahead and merge and start afresh with additional changes so that requesting a logger doesn't have side effects of logging elsewhere in the code.

PR #93, opened 2017-12-18
Independent log levels

### LINKED ISSUE #66: Questions/suggestions for logging module

weaverba137:
This PR closes #66:

* Calls to `get_logger()` return different loggers, depending on log level and other configuration details, so that a logger of level `DEBUG` does not interfere with a logger of level `INFO`.
* Creates a default logger with default level `INFO` at import time.  This is available as `desiutil.log.log`, *e.g.*:
```python
from desiutil.log import log
log.info('Information.')
```

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -7,9 +7,13 @@ Change Log
 
 * Enhance :mod:`desiutil.log` with a context manager (PR `#92`_), and
   change the way the log level is set.
+* Avoid logging interference with :func:`desiutil.log.get_logger` is called
+  with different log levels (PR `#93`_).
 * Use :mod:`unittest.mock` to increase test coverage.
 
 .. _`#92`: https://github.com/desihub/desiutil/pull/92
+.. _`#93`: https://github.com/desihub/desiutil/pull/93
+
 
 1.9.8 (2017-11-09)
 ------------------
--- a/py/desiutil/log.py
+++ b/py/desiutil/log.py
@@ -27,6 +27,11 @@ Examples
 
 Simplest possible use:
 
+>>> from desiutil.log import log
+>>> log.info('This is some information.')
+
+This is exactly equivalent to:
+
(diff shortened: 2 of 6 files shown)


weaverba137:
PS, `get_logger()` accepts case-insensitive strings in addition to integers, *i.e.*, these are all equivalent:
```python
log = get_logger(logging.DEBUG)
log = get_logger('DEBUG')
log = get_logger('debug')
```

sbailey:
Looks great.  Thanks.  OK to merge as soon as Travis tests re-pass.

PR #94, opened 2018-01-03
Add bspline to fitting

profxj:
Mainly a wrapper to scipy.interpolate routines.
Integrated within the func_fits(), iter_fit(),
and func_val() methods.

Includes new tests.

Passing on my machine but failing coverage
test on Travis for unclear reasons.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,6 +83,7 @@ before_install:
 install:
     - if [[ $MAIN_CMD == 'pycodestyle' ]]; then pip install pycodestyle; fi
     - if [[ $SETUP_CMD == build_sphinx* ]]; then pip install Sphinx; fi
+    - if [[ $SETUP_CMD == build_sphinx* ]]; then pip install scipy; fi  # Ugly kludge by JXP
     - if [[ $MAIN_CMD == 'coverage' ]]; then pip install coverage coveralls; fi
     - if [[ $SETUP_CMD == 'test' || $MAIN_CMD == 'coverage' ]]; then pip install -r requirements.txt; fi
 
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,7 @@ Change Log
 1.9.10 (unreleased)
 -------------------
 
-* Add bspline to funcfits
+* Add bspline to funcfits.  scipy is now a dependency
 
 1.9.9 (2017-12-20)
 ------------------
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,7 @@ Change Log
 1.9.10 (unreleased)
 -------------------
 
-* No changes yet.
+* Add bspline to funcfits
 
 1.9.9 (2017-12-20)
 ------------------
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -83,6 +83,7 @@ before_install:
 install:
     - if [[ $MAIN_CMD == 'pycodestyle' ]]; then pip install pycodestyle; fi
     - if [[ $SETUP_CMD == build_sphinx* ]]; then pip install Sphinx; fi
+    - if [[ $SETUP_CMD == build_sphinx* ]]; then pip install scipy; fi  # Ugly kludge by JXP
```
weaverba137: I'm pretty sure there is a more elegant way to do this, provided you actually understand how .travis.yml files work.  Is this something you would be able to figure out yourself?

doc/changes.rst
```diff
@@ -5,7 +5,7 @@ Change Log
 1.9.10 (unreleased)
 -------------------
 
-* No changes yet.
+* Add bspline to funcfits.  scipy is now a dependency
```
weaverba137: I have mixed feelings about adding scipy as an explicit dependency.  Will `bspline_fit()` be used by multiple DESI packages?  Because if it is being used by only one package, then it should be added to that package, not desiutil.

py/desiutil/test/test_funcfits.py
```diff
@@ -64,6 +69,18 @@ def test_cheby_fit(self):
         y2 = func_val(x2, dfit)
         np.testing.assert_allclose(y2[50], 0.99940823486206942)
 
+    def test_bspline_fit(self):
```
weaverba137: This test looks very simple.  Does it actually provide 100% coverage?

weaverba137:
The Travis failure was in the install stage, not an actual failure of the desiutil tests.

sbailey:
For due diligence, it is good to think twice before brining in a new dependency and prior to adding this, @profxj asked me if it was allowable to add scipy as a desiutil dependency.  Although I generally support minimizing dependencies, scipy is a fully approved core dependency for us and I told him to go ahead.  Travis and NERSC data transfer nodes are the only places that I can think of where we might use desiutil without needing scipy for other reasons, but it seems that adding scipy to both of those isn't particularly problematic.  Are there other cases and/or is there a specific problem with bringing in a scipy dependency?

How about moving the `from scipy import interpolate` into the various `bspline_*` functions such that it is only required if the code is actually run, and not for just importing the packages.

weaverba137:
I'm not actually that concerned *specifically* about the dependency on scipy.  I want to make certain that `bspline_fit()` is something that will actually be used by multiple DESI packages, and therefore merits being in desiutil (and therefore needs the scipy dependency).  If this is not the case, then we don't need to add `bspline_fit()` to desiutil and therefore *don't* need the scipy dependency.  No one has yet addressed this question.

julienguy:
I agree with Ben that only routines used in several other DESI packages should be added to desutil. I don't think it's the case with bspline_fit(). It would be better to have this in desispec in my opinion.

weaverba137:
I would like to close this PR and suggest that the functionality be added to desispec instead.  Maybe that's even already happened.

weaverba137:
I'm calling last call on keeping this open.  As previously stated, no one has made the case that this code *cannot* simply be added to desispec.  Unless someone objects in the next 1-2 days, I'm closing without merging.

weaverba137:
No objections were recorded, closing.

PR #101, opened 2018-03-12
Simplify desiInstall for use only at NERSC.

### LINKED ISSUE #96: Consider eliminating "cross-installs"
To reduce the complexity of desiInstall, consider eliminating "cross-installs".  A cross-install is where a product that contains no compiled code (*e.g.* only shell scripts and/or data) is installed only once and then symlinked into other install directories.  This has applied only to NERSC, where an install would be performed on edison, then symlinks would duplicate the installation for other NERSC hosts.

In practice, we don't actually use this that much, because we have very few products that fit into this category.  desimodules, desitree, and desiBackup are three products that do fit this category, but desimodules already requires special care in installation in the desiconda infrastructure.  So it doesn't seem worth it to carry forward the complexity of cross-installs just for a very small number of products.

@sbailey, please comment.

### LINKED ISSUE #97: Eliminate copying checked out code to final install directory
Historically as part of the `desiInstall` script, we *copied* the contents of the working/build directory to the final install directory, in addition to actually *installing* the code (`python setup.py install` / `make install`) in the same directory.

This isn't really necessary anymore, since we've moved necessary package data inside the Python packages themselves, so data gets installed along with the Python package.  However, a few packages, particularly desimodel, still need extra data that is downloaded to the *build* directory, then copied into place.

This issue will track any additional packages that need special data install handling, and the eventual removal of the code.  Other issues will track the modification of the data scripts to install directly to the install directory.

### LINKED ISSUE #99: Do we still need separate KNL support?
My understanding is that recent versions of desiconda are compiled in such a way that both haswell and knl are supported.  Thus it is no longer necessary to support separate cori and coriknl installations.  Can we confirm this?  It would simplify desiInstall a little bit.

### LINKED ISSUE #100: Ignore unimportant warning messages when installing specex
When compiling specex, warning like this are written to STDERR: `icpc: command line warning #10006: ignoring unknown option '-Wunused-value'`.  desiInstall interprets this output as a failure.

weaverba137:
This PR:

* Fixes #100.
* Fixes #99.
* Fixes #97.
* Fixes #96.

Still have a few more commits to add before merging...


weaverba137:
This is ready for review, but I want to run a few desiInstall tests before merging.

weaverba137:
Note to self: I still need to deal with suppressing specex warnings.

weaverba137:
Aside from desihub/fiberassign#103, I am satisfied that this update to desiInstall works.

PR #105, opened 2018-03-23
Python executable path

### LINKED ISSUE #104: installed  code shouldn't reference desiconda/current
A path problem for the 18.2 install, which we need to fix for 18.3:

```
[cori10 ~] source /project/projectdirs/desi/software/desi_environment_18.2.sh  
[cori10 ~] echo $DESICONDA
/global/common/software/desi/cori/desiconda/20180130-1.2.4-spec/conda
[cori10 ~] which desi_extract_spectra 
/global/common/software/desi/cori/desiconda/current/code/desispec/0.18.0/bin/desi_extract_spectra
[cori10 ~] head $(which desi_extract_spectra )
#!/global/common/software/desi/cori/desiconda/current/conda/bin/python
# EASY-INSTALL-SCRIPT: 'desispec==0.18.0','desi_extract_spectra'
__requires__ = 'desispec==0.18.0'
__import__('pkg_resources').run_script('desispec==0.18.0', 'desi_extract_spectra')
```

Note that those are in terms of desiconda/current instead of desiconda/20180130-1.2.4-spec .  Right now those are the same, but when we update which version of desiconda is "current" it will change the underlying desiconda version used when someone configures 18.2.  For 18.3, we should make sure that all the paths use desiconda/20180130-1.2.4-spec and don't rely upon "current".

As a procedure, I hope this is as simple as configuring desiconda/20180130-1.2.4-spec instead of desiconda/current before proceeding with installations.  It would be extra helpful if desiInstall could resolve symlinks into their final non-symlinked path; unfortunately `os.path.abspath` doesn't do that.

weaverba137:
This PR attempts to fix #104 at least for future installs by checking that during the bootstrap process, desiInstall is installed with an explicit desiconda version, not "current".

In theory, it would be possible to reject any attempt to use desiInstall if, internally, it cannot find an explicit desiconda version, but I have not implemented this.  I can if that is desired.  If we go down this road though, we may want to consider simply eliminating the 'current' symlink entirely, so that it is impossible not to choose an explicit version.

In addition, the desiutil.module file is added to the internal structure of the desiutil Python package so that it is available even if the etc/ directory is not installed, which it won't be, because we no longer copy the full package contents on install.

Finally, there is some clean-up of the package imports in `desiutil.install`.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -7,8 +7,11 @@ Change Log
 
 * Remove support for :command:`desiInstall` in environments other than
   NERSC (PR `#101`_).
+* Try as best as possible that Python executable scripts are installed with
+  an explicit desiconda version (PR `#105`_).
 
 .. _`#101`: https://github.com/desihub/desiutil/pull/101
+.. _`#105`: https://github.com/desihub/desiutil/pull/105
 
 1.9.9 (2017-12-20)
 ------------------
--- a/py/desiutil/data/desiutil.module
+++ b/py/desiutil/data/desiutil.module
@@ -1,0 +1,77 @@
+#%Module1.0
+# The first line of this file tells Modules that this is a module file.
+# DO NOT ALTER IT!
+#
+# ABOUT THIS FILE
+#
+# This file is designed to be processed by Python.  Specifically, this file
+# will be read into a string, and the .format() method will be applied to it.
+# This file is not a valid module file on its own.
+#
+# METADATA AND DOCUMENTATION SECTION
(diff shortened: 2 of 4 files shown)


weaverba137:
On edison, I hand-edited desiInstall, and installed desidatamodel 1.2.0.  The executable scripts in desidatamodel then had the correct Python paths.  So I think we'll be OK with just hand-editing desiInstall on cori & edison for the current desiconda version, then double-checking after the next bootstrap.

PR #111, opened 2018-04-19
update to more recent sklearn.mixture.GaussianMixture keywords

moustakas:
The sklearn data model changed recently.  This simple PR supports both old and newer versions of `sklearn.mixture.GaussianMixture` which is used in a variety of places in `desihub`.


sbailey:
sklearn has terrible datamodel consistency hygiene.  Thanks for contributing this workaround.

PR #116, opened 2018-08-27
Add SFD map code to the DESI utilities

geordie666:
This PR transfers the dust map code from `desitarget.mock.sfdmap` to `desiutil`. 

The context for porting this across is that the dust maps are now used in several places in `desitarget` (not just the mocks) and will likely also be needed for future updates to `desispec`. Some notes:

* Changes the default environment variable `SFD_DIR` to `DUST_DIR`
* Adds unit tests, which didn't exist in `desitarget`


moustakas:
This looks outstanding.  My only comment is that you consider renaming the `sfdmap` module something more generic like `dust`, so we can (in a separate PR) add one or more extinction curves, which will be needed in both `desitarget.mock` and `desispec` (to deredden spectra for Galactic extinction).

weaverba137:
There are some very significant problems with this PR.  But there is clearly still some active development going on, so please let me know when things are stable so I can review.

geordie666:
@weaverba137: Looks like the critical unit test are now passing, so feel free to elucidate the very significant problems.

geordie666:
Go for it with the review. That's a completely cosmetic change (hence the [ci skip]). I was just improving coding style, but it's nothing significant.

weaverba137:
A GitHub review needs to be attached to a particular commit.  If a new commit comes in while the review is going on, you have to refresh the files changed view and sometimes all your review comments can be lost.

py/desiutil/dust.py
```diff
@@ -0,0 +1,384 @@
+# Copied on Nov/20/2016 from https://github.com/kbarbary/sfdmap/ commit: bacdbbd
```
weaverba137: Desiutil already has a license, and we need to ensure that all code is compatible with that license.  Maybe we were sloppy about that when the code was in desitarget, but desiutil is typically held to a much higher standard.

py/desiutil/dust.py
```diff
@@ -0,0 +1,384 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+"""
+desiutil.dust
+=============
+
+Get E(B-V) values from the Schlegel, Finkbeiner & Davis (1998) dust map.
```
weaverba137: It would be helpful to add a link to the paper itself, or the abstract on ADS.

geordie666:
Noted. In future I won't make commits while reviews are pending, or I'll wait until I'm done with all commits before requesting a review. Hopefully that didn't happen in this case.

py/desiutil/dust.py
```diff
@@ -0,0 +1,384 @@
...
+try:
+    import fitsio
+    getdata = fitsio.read
+except ImportError:
+    try:
+        from astropy.io.fits import getdata
+    except ImportError:
+        raise ImportError("could not import fitsio or astropy.io.fits")
+
+__all__ = ['SFDMap', 'ebv']
```
weaverba137: No one in DESI should be doing `from desiutil.dust import *`, so this line is not necessary.

py/desiutil/dust.py
```diff
@@ -0,0 +1,384 @@
...
+__version__ = "0.1.0"
```
weaverba137: Desiutil already has a version, `desiutil.__version__`, so this really shouldn't be here.

py/desiutil/dust.py
```diff
@@ -0,0 +1,384 @@
...
+
+
+def _isiterable(obj):
```
weaverba137: There has to be a function in the Python standard library that does this.

py/desiutil/dust.py
```diff
@@ -0,0 +1,384 @@
...
+    try:
+        iter(obj)
+        return True
+    except TypeError:
+        return False
+
+# -----------------------------------------------------------------------------
+# Coordinate conversion
+#
+# astropy's coordinate conversions have a gigantic overhead of about
```
weaverba137: Considering that this code was written several years ago, this assertion should be verified.  If it is no longer true, we should use Astropy.

py/desiutil/dust.py
```diff
@@ -0,0 +1,384 @@
...
+# 20ms.  This kills performance in situations where you need to get a
+# single position at a time. We can do better (well under 0.1ms!) for
+# common static systems (IRCS, FK5J2000) by including a bit of code
+# from astropy.coordinates here.
+
+# Create rotation matrix about a given axis (x, y, z)
+
+def zrotmat(angle):
```
weaverba137: All functions need documentation strings, so they can be processed by Sphinx.

py/desiutil/dust.py
```diff
@@ -0,0 +1,384 @@
...
+def xrotmat(angle):
+    s = np.sin(angle)
+    c = np.cos(angle)
+    return np.array(((1, 0, 0),
+                     (0, c, s),
+                     (0, -s, c)))
+
+
+# constant ICRS --> FK5J2000 (See USNO Circular 179, section 3.5)
+try:
```
weaverba137: Why is this computation being performed at the module-level?  There is even a comment below that notes that!

py/desiutil/dust.py
```diff
@@ -0,0 +1,384 @@
...
+def _icrs_to_gal(lon, lat):
+    return cart2coords(np.dot(ICRS_TO_GAL, coords2cart(lon, lat)))
+
+
+def _fk5j2000_to_gal(lon, lat):
+    return cart2coords(np.dot(FK5J2000_TO_GAL, coords2cart(lon, lat)))
+
+
+# -----------------------------------------------------------------------------
+# bilinear interpolation (because scipy.ndimage.map_coordinates is really slow
```
weaverba137: This assertion should be justified, and if it is no longer true, we should use numpy/scipy/astropy.

py/desiutil/test/test_dust.py
```diff
@@ -0,0 +1,55 @@
...
+"""
+from __future__ import absolute_import, division, unicode_literals
+# The line above will help with 2to3 support.
+import unittest
+import os
+import numpy as np
+from .. import dust
+
+
+class TestDust(unittest.TestCase):
```
weaverba137: Assuming that `DUST_DIR` is defined, do the tests actually cover the entire module?  In other words, if we accept the module into desiutil, the total test coverage of desiutil should at least not decrease.

py/desiutil/dust.py
```diff
@@ -0,0 +1,384 @@
...
+
+Get E(B-V) values from the Schlegel, Finkbeiner & Davis (1998) dust map.
+"""
+
+import os
+
+import numpy as np
+
+# require a FITS reader of some sort.
+try:
```
weaverba137: This is not a common construction that I have seen in DESI software.  In fact, desiutil already requires astropy, so there is no reason to add an additional dependency on fitsio.

weaverba137:
An additional note as part of the previous review: as currently configured, no tests of this module are performed on Travis, because `DUST_DIR` is not *currently* defined in the Travis environment.  This also means that Travis will report to Coveralls that this module is not covered.

Data files are required to test this module, and would need to be brought into the Travis environment.  There are a few ways this could happen:

1. Download files as needed from a public site.
2. Include a "trimmed" version of the files in the desiutil test suite.
3. Use `unittest.mock` to simulate reading the files and inject appropriate data.

py/desiutil/dust.py
geordie666: I remain uncertain as to what to do about reusing code licensed in this manner. Now that you've brought this to my attention I certainly wouldn't merge this until we know the implications. I'm hoping @sbailey has some bright ideas.

py/desiutil/dust.py
geordie666: Done!

py/desiutil/dust.py
geordie666: I have removed this line.

py/desiutil/dust.py
geordie666: I've removed the version.

py/desiutil/dust.py
geordie666: I used np.atleast_1d() instead

py/desiutil/dust.py
geordie666: I have gone through the code and added sphinx-compliant doc strings. So, this should be solved, If unit tests pass.

py/desiutil/dust.py
geordie666: I have removed all of this code and use `astropy.coordinates.SkyCoord` objects instead.

py/desiutil/dust.py
geordie666: I haven't actually checked that (as I have a stable and working new version of the code and don't want to mess around with it much further), but I have removed that assertion. If you feel really strongly that we should use `scipy` then I'll assess this in detail.

py/desiutil/test/test_dust.py
geordie666: I've added a unit test and coverage is now ~94%.

py/desiutil/dust.py
geordie666: I have explicitly imported `astropy.io.fits` instead of this construction.

geordie666:
@weaverba137: Regarding your additional note...I chopped the dust maps down to the first 10 columns to reduce them in size to ~150K and use these chopped dust maps as the data for explicit unit tests. Your option (2), in other words.

weaverba137:
The code changes all look good to me know.  But let's make sure tests pass, & get feedback on license before merging.

geordie666:
I'm going to clean up my code style, so give me 10 minutes before anyone looks at this.

weaverba137:
OK, I'll double-check when you're done.

geordie666:
I'm done, now. There are a handful of other recommended code style alterations, but it's for a case where I think it's easier to read as-is.

weaverba137:
I'm trying to build the sphinx documentation on my own system and the dust module is not showing up.  Watch for additional commits to fix that.

moustakas:
@weaverba137 @geordie666 Is this ready to be merged?

If so, @geordie666 can you update `desitarget.mock.sfdmap` to use the dust utilities in `desiutil` and add a deprecation warning?

geordie666:
@moustakas: We're waiting for guidance on what to do about reconciling the license for this code with the DESI BSD license from @sbailey.

dkirkby:
+1 for adding at least one extinction curve.  Schlaflay 2016 is also an option (code [here](
https://faun.rc.fas.harvard.edu/eschlafly/apored/extcurve_s16.py)).  I doubt this would be a bottleneck for our applications, so straight numpy (w/o cython) should be sufficient if we impement this ourselves.

sbailey:
I'm joining this PR comment stream late, so apologies if I miss the point on something:

* I think license situation is fine: the main desiutil package is BSD licensed, and the dust.py file includes a comment that some portions were copied from https://github.com/kbarbary/sfdmap/ with the following MIT license, which seems to match both the spirit and the legalities of the open source licenses.  It would be slightly better if the comment specifically mentioned which functions / classes were copied (or is it the whole file?).

* Travis tests: looks like that has been resolved now, but in the future it is ok if Travis only runs a subset of the full tests if that is for a good reason, e.g. to avoid the overhead of large test files.  The full suite runs nightly at NERSC so we do get regular coverage.

* re: "Let's not pull in another package just to read the SFD maps".  I agree that we shouldn't pull in a new external dependency that has to be separately installed just to read the dust maps, but it is good to have a standardized convenience way to do that within our own code (e.g. here in desiutil rather than N>1 nearly identical implementations of the same thing in desitarget, desispec, desisurvey, desisim, ...)

weaverba137:
Thanks @sbailey.

> we do get regular coverage

Yes, but that coverage is not reported to coveralls.io.  Thus the coverage reported is lower than it should be, and there is not the immediate feedback of the Travis tests.

weaverba137:
PS, should we tag desiutil after this merge?

sbailey:
Agreed that having tests included in Travis is better when possible without heroic efforts.

I'll merge this now.

Please go ahead and make a new tag and install at NERSC.  I'll let @weaverba137 do that in order to pick the appropriate level of version number incrementing.

weaverba137:
1.9.12 is installed on cori and edison.

Issue #118, opened 2018-09-05
default mapdir assumed by dust.SFDMap won't jive with $DUST_DIR


moustakas:
In https://github.com/desihub/desimodules/issues/22 we added a new $DUST_DIR environment variable to point to `/global/project/projectdirs/cosmo/data/dust/v0_1`

However, `dust.SFDMap` needs/wants the full path to the dust maps, i.e., 
`/global/project/projectdirs/cosmo/data/dust/v0_1/maps`.

Specifically, at NERSC:
```
from desiutil.dust import SFDMap
ss = SFDMap()
ss.mapdir
  '/global/project/projectdirs/cosmo/data/dust/v0_1' # comes from $DUST_DIR
ss.ebv(120, 1.0)
[snip]
FileNotFoundError: [Errno 2] No such file or directory: '/global/project/projectdirs/cosmo/data/dust/v0_1/SFD_dust_4096_ngp.fits'
```
but
```
from desiutil.dust import SFDMap
ss = SFDMap(mapdir='/global/project/projectdirs/cosmo/data/dust/v0_1/maps')
ss.ebv(120, 1.0)
```
works.

So I think that `SFDMap.__init__` just needs to append the `maps` subdirectory when evaluating `$DUST_DIR` (which I *don't* think we should change).

weaverba137:
Fixed by #119.

PR #119, opened 2018-09-05
adding the maps subdirectory to the dustdir variable

geordie666:
This PR updates the default directory used by `desiutil.dust` to add the `/maps` subdirectory to the `$DUST_DIR` environment variable.

### DIFF
--- a/py/desiutil/dust.py
+++ b/py/desiutil/dust.py
@@ -167,11 +167,12 @@ class SFDMap(object):
 
     Parameters
     ----------
-    mapdir : :class:`str`, optional, defaults to :envvar:`DUST_DIR`.
+    mapdir : :class:`str`, optional, defaults to :envvar:`DUST_DIR`+``/maps``.
         Directory in which to find dust map FITS images, named
         ``SFD_dust_4096_ngp.fits`` and ``SFD_dust_4096_sgp.fits``.
-        If not specified, the value of the :envvar:`DUST_DIR` environment
-        variable is used, otherwise an empty string is used.
+        If not specified, the map directory is derived from the value of 
+        the :envvar:`DUST_DIR` environment variable, otherwise an empty 
+        string is used.
     north, south : :class:`str`, optional
         Names of north and south galactic pole FITS files. Defaults are
         ``SFD_dust_4096_ngp.fits`` and ``SFD_dust_4096_sgp.fits``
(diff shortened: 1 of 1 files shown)


geordie666:
@moustakas: is this what you had in mind to fix https://github.com/desihub/desiutil/issues/118? If so, ask @weaverba137 for a review prior to any merge.

py/desiutil/dust.py
```diff
@@ -189,7 +190,8 @@ def __init__(self, mapdir=None, north="SFD_dust_4096_ngp.fits",
                  south="SFD_dust_4096_sgp.fits", scaling=1.):
 
         if mapdir is None:
-            mapdir = os.environ.get('DUST_DIR', '')
+            dustdir = os.environ.get('DUST_DIR', '')
```
weaverba137: So if `DUST_DIR` is not set, `dustdir = ''`, and `mapdir = '/maps'`.  I think it would be better to raise an exception saying that the maps can't be found, rather than wait until a somewhat obscure error is raised when looking for fits files in a directory that doesn't exist.

py/desiutil/dust.py
```diff
@@ -189,7 +190,8 @@ def __init__(self, mapdir=None, north="SFD_dust_4096_ngp.fits",
...
+            mapdir = "{}/maps".format(dustdir)
```
weaverba137: Use `mapdir = os.path.join(dustdir, 'maps')`.

geordie666:
Oops, @weaverba137 I just found a minor typo. Please hold off any further review.

geordie666:
OK, @weaverba137, I'm done with my review response. Thanks.

py/desiutil/dust.py
```diff
@@ -189,7 +192,17 @@ def __init__(self, mapdir=None, north="SFD_dust_4096_ngp.fits",
                  south="SFD_dust_4096_sgp.fits", scaling=1.):
 
         if mapdir is None:
-            mapdir = os.environ.get('DUST_DIR', '')
+            dustdir = os.environ.get('DUST_DIR')
+            if dustdir is None:
+                log.critical('Pass mapdir or set $DUST_DIR')
+                raise IOError
```
weaverba137: Typically when raising an exception, you pass a message to it.  In this case, the same message as the log message will do.

Also `IOError` might not be ideal here because the exception hierarchy is different between 2 & 3.  I recommend carefully reading https://docs.python.org/3/library/exceptions.html to see if that's what you really want.  For example, `ValueError` might be just as appropriate.

Please also make sure that the unit tests actually test the case where an invalid value is passed during initialization.

geordie666: I now raise a ValueError with a message, and I've added a unit test to check that ValueErrors are thrown appropriately when `DUST_DIR` doesn't exist or when a meaningless map directory is passed.

weaverba137:
Looks good.  Merge when tests are done.  I will tag after the merge.

weaverba137:
desiutil/1.9.13 is installed on cori and edison.

PR #121, opened 2018-09-20
Python 3

weaverba137:
This PR eliminates all support for Python 2.  This does not need to be merged immediately; it is more of an exercise in what we can we do when we drop Python 2 support.


sbailey:
What is the motivation for this update?  It appears that it drops support for py2 but I don't see what the actual benefit is.  Does it let us do something that wasn't possible before?  It's fine if we want to add a new piece of code that for whatever reason only works with py3, but going back and stripping out py2 support feels like a distraction.

weaverba137:
As I mentioned in the description, this is an exercise.  I found quite a bit of code that is completely unnecessary, and might even be confusing in the future, given that we won't be using Python 2 anyway.

In addition, Astropy has taken the proactive step of removing Python-2-compatible code, and I wanted to see what that would look like.

sbailey:
OK as an exercise, but please don't spend the time to do this for every other repo right now.  I'm also inclined to *not* merge this in desiutil right now, because it is at the bottom of our dependency chain and if desiutil breaks py2, then effectively that breaks py2 for everything.  We don't need to spend effort maintaining py2 if some new piece of code happens to not work with it, but let's not spend our time purposefully breaking things that still work.  OK to leave branch open for merging someday, though perhaps closing the PR (at minimum I'm going to remove it from the 18.9 critical path).

weaverba137:
Agreed.  In the meantime, I'm going to remove .gitlab-ci.yml from master.

weaverba137:
@sbailey , could you revisit this when you get a chance?  This has been hanging for almost a year. At this point, I think anyone still trying to use desiutil with Python 2 should be forced to upgrade.

weaverba137:
Thank you!

Issue #114, opened 2018-08-18
change declination range of equatorial/basemap projections?


moustakas:
@dkirkby In `plots.plot_sky_binned` I'd like to change the declination range of the map to, e.g., to zoom into the southern or northern cap region.  I played around with the `lat_[0,1,2]` and `llcrnrlat` and `urcrnrlat` arguments to the `Basemap` class (around line 460 in `init_sky`) but didn't have much luck.

This request may not be possible but I thought I'd ask.

weaverba137:
Pinging @dkirkby.  This question/issue is for you.

weaverba137:
It is not possible to change axis limits on built-in matplotlib projections.  Since we're trying to eliminate basemap and related external dependencies, we're not going to support that *in desiutils*.  You are welcome to share other solutions you find with the collaboration.

PR #124, opened 2018-10-18
Add ecliptic plane to all-sky plots

dkirkby:
Draw the ecliptic in `plots.init_sky()`.  Shown as a dotted red line by default, but can use a different color or turned off using the `ecliptic_color` parameter.

expfac_airmass_ha0

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -10,7 +10,7 @@ Change Log
 1.9.15 (unreleased)
 -------------------
 
-* No changes yet.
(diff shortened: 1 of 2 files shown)


weaverba137:
This looks fine, but while I've got your attention, is anything happening with #114?

weaverba137:
Sorry this has been dangling for a while.  If there are no objections, I'll merge this soon.

weaverba137:
"Soon" = after I got back from Chile.

PR #125, opened 2018-10-30
Permission unlock

### LINKED ISSUE #123: desiInstall can't alter .version files with new permissions settings.
So, I tagged 1.9.14, and went to install it with `desiInstall -d desiutil 1.9.14`.  In the process of restricting write access, even to the desi user, I also made .version files read-only.  However, `desiInstall` expects to be able to write to such files.  For now, I am going to restore user-read to all .version files in 20180709-1.2.6-spec.

weaverba137:
This PR fixes #123.  In so doing:

* Permission changes on Module files are moved into the `modules.py` file.
* A new context manager, `unlock_file()`, temporarily grants user-write permission on a read-only file.  This might be useful in other code, so it's available in `desiutil.io`.  It is a drop-in replacement for the Python built-in `open()`.


weaverba137:
In my first attempt to code this up, I didn't consider what happens when the file doesn't already exist, in which case, it doesn't have permissions that can be restored.  There are at least two possibilities then:

1. If the file doesn't already exist, `unlock_file()` just opens the file and closes it, but doesn't do anything else.  That is, the user would have to explicitly turn off write permission after.
2. If the file doesn't already exist, turn off all write permission bits after the file is closed.

sbailey:
Use of `unlock_file` in code would imply that the user expects that the file is write-locked, and is left in a write-locked state when done.  Thus I would vote option (2): "If the file doesn't already exist, turn off all write permission bits after the file is closed."

weaverba137:
OK, I've implemented it such that all write bits are turned off after using unlock_file, even if the file previously did exist and had some write permissions set.

sbailey:
Hmm, this is a slippery slope.  I would have expected this to temporarily enable write access as needed, and then:
1. if the file exists, leave the permissions alone when done (even if it had write permission to start with)
2. if the file doesn't exist, leave user and group read and no write when done

weaverba137:
OK, but you're directly contradicting yourself:
> the user expects that the file is write-locked, and is left in a write-locked state when done.
versus
> if the file exists, leave the permissions alone when done (even if it had write permission to start with)

In addition,
> if the file doesn't exist, leave user and group read and no write when done
I think that's dangerous to be so exactly prescriptive of the permissions.  It would be much easier to say "remove all write permissions; leave any other bits alone, as whatever `open(filename, 'w')` would have done".

sbailey:
We seem to always get ourselves in a mess over file permissions.  After letting this simmer for awhile:
* I like how you originally implemented it, where if the file exists this code restores the permissions to whatever they were without asking questions
* If the file doesn't exist, I'd suggest leaving it without write permission, but if you feel strongly the other way, that's fine too

Note: using this function implies (to me) that the user expects that the file is read-only and they want to temporarily override that.  But it seems like a tricky side effect to have this code *enforce* the user's expectation on the file even if that expectation was wrong.  Thus I advocate leaving the permissions alone for pre-existing files.

Side note: even though this PR has been dangling for 2 weeks, I'll note that we had 2 reviews, Halloween, and Veterans Day in there (not to mention an observing run for @weaverba137 and raging wildfires and smoke up here...)

weaverba137:
@sbailey, thanks for that, I'll probably finish this off today.

weaverba137:
I think this is ready to merge, last call for review.

PR #128, opened 2018-11-26
add desiutil.dust.ext_odonnell and ext_ccm from desispec

sbailey:
This PR moves `ext_odonnell` and `ext_ccm` from `desispec.dustextinction` into `desiutil.dust` (which already had SFD E(B-V) code) in preparation for both desitarget and desispec using these extinction laws.

I think this is what we agreed to on the last DESI data telecon, but mentioning @julienguy  @moustakas @geordie666 @weaverba137 for comment if they disagree.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -10,10 +10,13 @@ Change Log
 1.9.15 (unreleased)
 -------------------
 
+* Add :func:`desiutil.dust.ext_odonnell` and :func:`desiutil.dust.ext_ccm`
+  originally from desispec (PR `#128`).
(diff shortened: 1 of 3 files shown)


doc/changes.rst
```diff
@@ -10,10 +10,13 @@ Change Log
 1.9.15 (unreleased)
 -------------------
 
+* Add :func:`desiutil.dust.ext_odonnell` and :func:`desiutil.dust.ext_ccm`
+  originally from desispec (PR `#128`).
```
weaverba137: Missing a `_` here.

julienguy:
ok for me

PR #141, opened 2020-04-15
Replace basemap

### LINKED ISSUE #132: Replace basemap
This issue is to replace the use of basemap in this package with a pure-matplotlib implementation of the map projection, to resolve #129.

Note that matplotlib already has built-in support for some  all-sky projections:
```
matplotlib.projections.get_projection_names()
['aitoff', 'hammer', 'lambert', 'mollweide', 'polar', 'rectilinear']
```

weaverba137:
This PR fixes #132.  Only built-in matplotlib routines are used.  The example notebook, SkyMapExamples.ipynb has been rewritten to demonstrate all of the old functionality without basemap.

There is also a new notebook, SVFields.ipynb that might be useful.


weaverba137:
Yes, I figured there would be some tests to patch up.

One important change is that the `basemap` keyword is replaced by `ax`.  But that only applies if you have an existing axis object and want to overplot something.

Otherwise I am pretty sure that it is API-compatible, although there may have been some changes to the other keywords accepted by `init_sky()`.  If you compare the example notebook, the calling signatures of the functions didn't change.  I only cleaned up the presentation there.

sbailey:
@geordie666 can you confirm that target selection QA works with this branch?  Or if you don't have time today punt it back to us for someone else to try?  I'd like to include this in the 20.4 software release (which is slightly delayed for other reasons) but I want to confirm that it doesn't break desitarget first if that can be done in time.

geordie666:
I'm on it...give me a few hours.

weaverba137:
In particular, please re-enable the disabled tests as described in desihub/desitarget#578.

geordie666:
I had to make some (very minor) changes to integrate desitarget QA with this branch. Target selection QA will work with this branch once we merge [desitarget #605](https://github.com/desihub/desitarget/pull/605). 

Also, reintroducing the now-3-months-old desitarget QA unit tests broke end-to-end integration (as detailed in [desitarget #605](https://github.com/desihub/desitarget/pull/605)) so that will require some further minor fixes.

PR #151, opened 2020-09-09
add Timer class to standardize algorithm timing reports

sbailey:
This PR adds a lightweight `desiutil.timer.Timer` utility class to standardize timing reports of pipeline steps.  It is intended for steps that take seconds to minutes, not for high-resolution profile timing of N>>1 functions.  e.g. in desispec pipeline steps I plan to track python startup, initial input, algorithm, and final output times.

Example usage with either `Timer.start(name)` and `Timer.stop(name)` or with a context manager `Timer.time(name)`:
```python
import time
from desiutil.timer import Timer
t = Timer()

# Start a named timer
t.start('blat')
# do something...
time.sleep(0.1)

# Use a context manager to time a small snippet
with t.time('input'):
    # stuff = io.read(filename)
    time.sleep(0.1)

# do some more stuff
time.sleep(0.1)

# stop the blat timer
t.stop('blat')

# Get a json-parseable report
print(t.report())
```
Note that named timers can be nested and interleaved.  If run as a script `blat.py` this would produce:
```
TIMER-START:blat.py:6:<module>: Starting blat at 2020-09-09T12:05:36.884383
TIMER-START:blat.py:11:<module>: Starting input at 2020-09-09T12:05:36.989992
TIMER-STOP:blat.py:13:<module>: Stopping input at 2020-09-09T12:05:37.091668; duration 0.10 seconds
TIMER-STOP:blat.py:19:<module>: Stopping blat at 2020-09-09T12:05:37.194143; duration 0.31 seconds
{
  "blat": {
    "start": "2020-09-09T12:05:36.884723",
    "stop": "2020-09-09T12:05:37.194197",
    "duration": 0.30947399139404297
  },
  "input": {
    "start": "2020-09-09T12:05:36.991274",
    "stop": "2020-09-09T12:05:37.091691",
    "duration": 0.1004171371459961
  }
}
```

Implementation detail: this does *not* use the `desiutil.log` logger, since I want the filename and line numbers to match the caller of the timer, not the line numbers within the `timer.py`, which is what would appear if `Timer` was calling `log.info()` etc.

The code is also intended to be robust to misuse, e.g. re-starting a named timer that is already running, stopping a timer that was never started, or stopping a timer twice.  These produce warning/error messages, but don't crash.

Review requested from @weaverba137 as the gatekeeper of desiutil, and comments welcome from @akremin @julienguy @dmargala @lastephey or any others who have recently been thinking about what would be useful for more standardized timing reports.

### DIFF
--- a/py/desiutil/timer.py
+++ b/py/desiutil/timer.py
@@ -1,0 +1,125 @@
+"""
+timer: standardizing reporting of algorithm and I/O timing
+
+The `timer.Timer` class is intended for reporting timing of events that
+take seconds to minutes; it is not intended for detailed profiling.  Example::
+
+    from desiutil.timer import Timer
+    t = Timer()
+    t.start('blat')
+
+    #- Use context manager for timing simple steps, e.g. one-liners
+    with t.time('blat.input'):
+        stuff = io.read(filename)
+
+    #- Or use full start/stop if you prefer
+    t.start('blat.algorithm')
+    results = calculate(stuff)
+    t.stop('blat.algorithm')
+
+    with t.time('blat.output'):
+        io.write(results)
+
+    #- Stop outer timer; Note that named timers can be nested or interleaved
+    t.stop('blat')
+
+    #- Print a json report of the timing
+    print(t.report())
+"""
+
+import traceback
+import time
+import datetime
+import json
+import copy
+import os.path
+from contextlib import contextmanager
+
+class Timer(object):
+    """
+    A basic timer class for standardizing reporting of algorithm and I/O timing
+    
+    TIMER:<START|STOP>:<filename>:<lineno>:<funcname>: <message>
+    """
+    
+    def __init__(self):
+        self.timers = dict()
+    
+    def _prefix(self, step):
+        """
+        Return standard prefix string for timer reporting
+        
+        Args:
+            step (str): timing step, e.g. "START" or "STOP"
+        """
+        stack = traceback.extract_stack()
+        #- Walk backwards in stack to find first caller not from this file
+        #- and not contextlib.py
+        #- (exact index depends on whether context manager was used or not)
+        thisfile = os.path.normpath(__file__)
+        for caller in stack[-1::-1]:
+            if os.path.normpath(caller.filename) != thisfile and \
+               os.path.basename(caller.filename) != 'contextlib.py':
+                break
+        
+        filename = os.path.basename(caller.filename)
+        return f"TIMER-{step}:{filename}:{caller.lineno}:{caller.name}:"
+    
+    def start(self, name):
+        """Start a timer `name` (str)"""
+        # timestamp = datetime.datetime.now().replace(microsecond=0).isoformat()        
+        timestamp = datetime.datetime.now().isoformat()        
+        if name in self.timers:
+            print(self._prefix('WARNING'), f'Restarting {name} at {timestamp}')
+        
+        print(self._prefix('START'), f'Starting {name} at {timestamp}')
+        self.timers[name] = dict(start=time.time())
+    
+    def stop(self, name):
+        """Stop timer `name` (str)"""
+        #- non-fatal ERROR: trying to stop a timer that wasn't started
+        timestamp = datetime.datetime.now().isoformat()        
+        if name not in self.timers:
+            print(self._prefix('ERROR'), f'Tried to stop non-existent timer {name} at {timestamp}')
+            return -1.0
+
+        #- WARNING: resetting the stop time of a timer that was already stopped
+        if 'stop' in self.timers[name]:
+            print(self._prefix('WARNING'), f'Resetting stop time of {name} at {timestamp}')
+        
+        #- All clear; proceed
+        self.timers[name]['stop'] = time.time()
+        dt = self.timers[name]['stop'] - self.timers[name]['start']
+        self.timers[name]['duration'] = dt
+        print(self._prefix('STOP'), f'Stopping {name} at {timestamp}; duration {dt:.2f} seconds')
+        return dt
+
+    @contextmanager
+    def time(self, name):
+        self.start(name)
+        try:
+            yield
+        finally:
+            return self.stop(name)
+
+    def report(self):
+        """
+        Return a json dump of self.timers, with start/stop as ISO-8601 strings
+        
+        Implicitly stops any timers that are still running
+        """
+        #- First stop any running timers
+        for name in self.timers:
+            if 'stop' not in self.timers[name]:
+                self.stop(name)
+        
+        #- Now replace time-since-epoch float with ISO-8601 str using copy
+        t = copy.deepcopy(self.timers)
+        for name in t:            
+            for key in ['start', 'stop']:
+                tx = datetime.datetime.fromtimestamp(t[name][key]).isoformat()
+                t[name][key] = tx
+
+        return json.dumps(t, indent=2)
+        
+
--- a/py/desiutil/timer.py
+++ b/py/desiutil/timer.py
@@ -1,5 +1,11 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
 """
-timer: standardizing reporting of algorithm and I/O timing
+==============
+desiutil.timer
+==============
+
+Provides `Timer` class to standardize reporting of algorithm and I/O timing.
 
 The `timer.Timer` class is intended for reporting timing of events that
 take seconds to minutes; it is not intended for detailed profiling.  Example::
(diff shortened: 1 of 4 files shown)


weaverba137:
Unit test coverage for `timer.py` is 100% so thank you for starting from that.

py/desiutil/timer.py
```diff
@@ -0,0 +1,125 @@
...
+        
+        filename = os.path.basename(caller.filename)
+        return f"TIMER-{step}:{filename}:{caller.lineno}:{caller.name}:"
+    
+    def start(self, name):
+        """Start a timer `name` (str)"""
+        # timestamp = datetime.datetime.now().replace(microsecond=0).isoformat()        
+        timestamp = datetime.datetime.now().isoformat()        
+        if name in self.timers:
+            print(self._prefix('WARNING'), f'Restarting {name} at {timestamp}')
```
weaverba137, line 73: I'm very concerned that you are using bare `print()` statements here.  If the intention is to *only* use this in a laptop command-line or Jupyter notebook environment, maybe that's fine.  But it is not clear from the PR comments or the documentation in the module itself what the intended use environment is.

You do say that you're deliberately not using `desiutil.log`, but what if you need the timing in something that is sending output to a log file?  Yes `desiutil.log` writes to stdout, but that is also often overridden.  Furthermore `desiutil.log` may buffer output differently even when writing to stdout, resulting in log messages and timing messages being written out of order.

py/desiutil/timer.py
```diff
@@ -0,0 +1,125 @@
...
+        
+        print(self._prefix('START'), f'Starting {name} at {timestamp}')
+        self.timers[name] = dict(start=time.time())
+    
+    def stop(self, name):
+        """Stop timer `name` (str)"""
+        #- non-fatal ERROR: trying to stop a timer that wasn't started
+        timestamp = datetime.datetime.now().isoformat()        
+        if name not in self.timers:
+            print(self._prefix('ERROR'), f'Tried to stop non-existent timer {name} at {timestamp}')
```
weaverba137, line 83: I think it would be better to raise an exception here.  If that is not possible, at least return a `TimeDelta` rather than some other type.

py/desiutil/timer.py
```diff
@@ -0,0 +1,125 @@
...
+                self.stop(name)
+        
+        #- Now replace time-since-epoch float with ISO-8601 str using copy
+        t = copy.deepcopy(self.timers)
+        for name in t:            
+            for key in ['start', 'stop']:
+                tx = datetime.datetime.fromtimestamp(t[name][key]).isoformat()
+                t[name][key] = tx
+
+        return json.dumps(t, indent=2)
```
weaverba137, line 123: It might be nice to have another utility function that converts the `datetime` data to strings and returns that dictionary, then have `report` just do `json.dumps`.  Otherwise if the user needs a Python dictionary, the user would have to convert JSON *back* to a Python `dict()`.

Alternatively, document examples of user access to the `timers` attribute, since it is not deliberately hidden.

py/desiutil/timer.py
```diff
@@ -0,0 +1,125 @@
...
+        # timestamp = datetime.datetime.now().replace(microsecond=0).isoformat()        
+        timestamp = datetime.datetime.now().isoformat()        
+        if name in self.timers:
+            print(self._prefix('WARNING'), f'Restarting {name} at {timestamp}')
+        
+        print(self._prefix('START'), f'Starting {name} at {timestamp}')
+        self.timers[name] = dict(start=time.time())
+    
+    def stop(self, name):
+        """Stop timer `name` (str)"""
```
weaverba137, line 79: Document that `stop()` returns a value.  See also next comment.

weaverba137 (changes requested):
This looks very interesting, but there are a number of issues to look at in practice.  Many of the issues can be addressed with documentation.

py/desiutil/timer.py
weaverba137: One more thought:  if Jupyter notebooks *are* an intended environment for this, document why exactly one would want to use this as opposed to Jupyter built-ins like `%time` and `%%time`.

tskisner:
Probably you don't care, but in case some other Timer implementations are interesting to you:

In TOAST, both individual timers:

https://github.com/hpc4cmb/toast/blob/master/src/toast/_libtoast_sys.cpp#L98

And a singleton class which acts as a per-process registry of global timers that can be started / stopped anywhere in the code:

https://github.com/hpc4cmb/toast/blob/master/src/toast/_libtoast_sys.cpp#L258

The test cases for these:

https://github.com/hpc4cmb/toast/blob/master/src/toast/tests/timing.py

A function decorator which accumulates the aggregrate times for a function:

https://github.com/hpc4cmb/toast/blob/master/src/toast/timing.py#L20

Also in that same file are functions that gather global timers from all processes and dump the min/max/mean/median to a CSV file.

As you can see from the links, these are implemented in C++ with pybind11 bindings (in order to unify the timing of both compiled and python code), but you could easily implement those in pure python if you wanted.

py/desiutil/timer.py
sbailey: Do you have suggestions for how to get this to play nicely with a logger while still having it print the filenames and line numbers for the code that called the timer, instead of for the line numbers within `timer.py` that call the logger?

e.g. if I have a script `blat.py`:
```python
import time
from desiutil.timer import Timer
t = Timer()
t.start('blat')
time.sleep(0.1)
t.stop('blat')
```
The current timer code produces:
```
TIMER-START:blat.py:4:<module>: Starting blat at 2020-09-09T14:25:54.262900
TIMER-STOP:blat.py:6:<module>: Stopping blat at 2020-09-09T14:25:54.364106; duration 0.10 seconds
```
But if I replace the bare print statements with log.info calls, I get:
```
INFO:timer.py:84:start: Starting blat at 2020-09-09T14:25:04.315599
INFO:timer.py:106:stop: Stopping blat at 2020-09-09T14:25:04.420473; duration 0.10 seconds
```
i.e. it preserves the fact that timers are starting and stopping, but not what file/lineno started/stopped them, which is what I also want in the messages, but I'm not sure how to do that with a logger.

py/desiutil/timer.py
```diff
@@ -0,0 +1,125 @@
+"""
```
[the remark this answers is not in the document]
weaverba137: Fixed in a4c02fb

py/desiutil/timer.py
weaverba137: I don't have any off-the-cuff suggestions.  Answering that would require some in-depth research, not something you can do in a few minutes.

Real-world testing of the existing code might prove to be more important here.  For example, if this turns out this is fundamentally broken in a multiprocessing/mpi/openmp/... environment, then maybe it's OK to keep the print statements, since it would only ever work in a small command-line script.

py/desiutil/timer.py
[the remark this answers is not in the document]
sbailey: done

py/desiutil/timer.py
sbailey: done (both separating out the conversion function, and documenting `Timer.timers`)

py/desiutil/timer.py
sbailey: done

py/desiutil/timer.py
sbailey: I documented the philosophy more, but I do want to stick with the safety that adding timer info shouldn't be able to crash user code, even if used incorrectly.  This is an exception to my normal guideline of fail-early fail-often.

Also, internally the timestamps are tracked as `time.time()` seconds since epoch, not `datetime` instances, so a correct call to `stop` returns a float difference in seconds not a `timedelta`, so I kept the incorrect call to stop consistent with that by returning a float.  I updated the documentation to be clear about what is returned.

sbailey:
Thanks for the review @weaverba137 .  I think I have addressed all the suggestions (though in some cases adding documentation but not changing implementation), except the outstanding issue of whether there is a way to make this more integrated with the logging framework while still preserving line numbers of the calling code, not the timer code.  Ideas welcome.  If it comes down to a tradeoff, I'm inclined to stick with the bare bring statements with the line numbers I want, and then revisit that if they become problematic in practice.

The motivating use-case for this Timer class is augmenting each step of the spectro pipeline to get clear, standardized, and easily parseable timing info separated by input, calculation, and output.

Thanks for those pointers @tskisner.  I certainly took inspiration from some of your earlier timer work, but decided to implement something lighter weight here as a starter and then add more features later if/when we need them (e.g. I cut myself off from adding a lap timer for loops, and min/median/mean/max for this version).

weaverba137 (approved):
Please update the change log, but go ahead and merge so we can start testing.

sbailey:
changes.rst updated.

For the record, two last minute changes:
  * factored out `Timer.stopall()` to stop any timers that had not yet been individually stopped
     (previously that was hidden within `Timer.report()` which also stopped any running timers).
  * small update to make ISO8601 timestamp in log messages identical to what ends up in the report.

weaverba137:
Sounds good.  Merge when ready.

PR #152, opened 2020-09-25
more Timer functionality

sbailey:
This PR adds several features that I wanted when actually putting it into use in desispec:
  * silent option: record timing, but don't print log messages, e.g. to record timing for multiple MPI ranks without having every rank blasting timing log messages.
  * `calculate_stats` (like Ted's example from CMB-land suggested in the original Timer PR #151): combine multiple Timers into a report with min/max/mean/median of start/stop/duration.
  * options to `start` and `stop` to set the time to use instead of `time.time()`, useful for recording timing of events that may have occurred before the timer was created, e.g. python startup and imports.
  * `Timer.cancel` to discard a named timer that had been started.

I'll submit a companion desispec PR that uses these features, but this PR will need to be reviewed and merged before that one can be finalized.

### DIFF
--- a/py/desiutil/test/test_timer.py
+++ b/py/desiutil/test/test_timer.py
@@ -126,10 +126,10 @@ class TestTimer(unittest.TestCase):
     @unittest.skipUnless(dateutil_installed, "dateutil not installed")
     def test_parsetime_dateutil(self):
         """If dateutil installed, test fancy date string parsing"""
-        t0 = parsetime("Tue Sep 22 13:42:26 PDT 2020")
+        # t0 = parsetime("Tue Sep 22 13:42:26 PDT 2020")
+        # self.assertAlmostEqual(t0, 1600807346.0)
         t1 = parsetime("2020-09-22T13:42:26-07:00")
-        self.assertAlmostEqual(t0, t1)
-        self.assertAlmostEqual(t0, 1600807346.0)
+        self.assertAlmostEqual(t1, 1600807346.0)
 
         with self.assertRaises(ValueError):
             t2 = parsetime("My Birthday")
@@ -137,8 +137,8 @@ class TestTimer(unittest.TestCase):
     @unittest.skipIf(dateutil_installed, "dateutil installed")
     def test_parsetime_no_dateutil(self):
         """If dateutil not installed, confirm failure modes"""
-        with unittest.assertRaises(ValueError):
-            t0 = parsetime("Tue Sep 22 13:42:26 PDT 2020")
+        # with unittest.assertRaises(ValueError):
+        #     t0 = parsetime("Tue Sep 22 13:42:26 PDT 2020")
         with unittest.assertRaises(ValueError):
             t1 = parsetime("2020-09-22T13:42:26-07:00")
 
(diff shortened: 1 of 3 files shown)


sbailey:
Travis tests are catching a timezone parsing issue that doesn't arise at NERSC or on my laptop.  I'll investigate what the difference is, and back off on fancy date string parsing if needed.  There might be some Travis churn as I sort this out.

weaverba137:
You're using `pytz` right?

sbailey:
No I wasn't using pytz (only vaguely aware of it...)  Maybe for now I will just drop support for the fully generic Unix `date` output and require `date +%s` or ISO-8601 (which itself has lots of options...).  I can return to adding vanilla `date` support if the pain of not having it exceeds the pain of implementing it.

weaverba137:
OK, ready for me to take a look?

sbailey:
Yes, please.  Thanks.

py/desiutil/test/test_timer.py
```diff
@@ -85,6 +106,80 @@ def test_timer_misuse(self):
...
+        self.assertAlmostEqual(t0, t2)
+        self.assertAlmostEqual(t0, t3)
+
+        timer = Timer()
+        timer.start('blat', starttime=1600807346)
+        timer.stop('blat', stoptime=1600807346+2)
+        self.assertAlmostEqual(timer.timers['blat']['start'], 1600807346.0)
+        self.assertAlmostEqual(timer.timers['blat']['duration'], 2)
+
+    @unittest.skipUnless(dateutil_installed, "dateutil not installed")
```
weaverba137, line 126: Might want to add a comment that fancy date parsing is currently disabled in the module being tested, but may be reimplemented in the future.

py/desiutil/timer.py
```diff
@@ -149,6 +180,17 @@ def stopall(self):
             if 'stop' not in self.timers[name]:
                 self.stop(name)
 
+    def cancel(self, name):
+        """Cancel timer `name` and remove from timing log"""
+        t1 = time.time()
+        isotime = datetime.datetime.fromtimestamp(t1).isoformat()
+        if name in self.timers:
+            dt = t1 - self.timers[name]['start']
+            print(self._prefix('CANCEL'), f'Canceling timer {name} at {isotime} after {dt:.2f} seconds')
```
weaverba137, line 189: Use `self._print()` instead?

weaverba137 (approved):
Overall looks fine, but please note the comments.  Most importantly, make sure `self._print()` is used consistently throughout the `Timer` class.

weaverba137:
PS, please update the change log.

py/desiutil/timer.py
sbailey: Good catch; I missed this one.  Updated to use `self._print`

py/desiutil/test/test_timer.py
sbailey: comment added about ISO-8601 vs. fancier date parsing in the future.

weaverba137:
OK to merge assuming tests pass.

PR #153, opened 2020-11-20
Work on moving stdout / stderr redirection from desispec to desiutil

tskisner:
The unit tests are not yet working, since redirecting stdout actually breaks the unit test runner (which has a persistent handle to sys.stdout).  I am working on a unit test that runs the redirection in a subprocess.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -95,7 +95,7 @@ install:
     - pip install -U pip
     - if [[ "${MAIN_CMD}" == pycodestyle ]]; then pip install pycodestyle; fi
     - if [[ "${MAIN_CMD}" == sphinx-build* ]]; then pip install Sphinx; fi
-    - if [[ "${MAIN_CMD}" == pytest ]]; then pip install -r requirements.txt; fi
+    - if [[ "${MAIN_CMD}" == pytest ]]; then pip install -r requirements.txt; pip install . ; fi
     - if [[ "${SETUP_CMD}" == '--cov' ]]; then pip install pytest-cov coveralls; fi
     # - if [[ "${MAIN_CMD}" == python* || "${MAIN_CMD}" == coverage* ]]; then pip install -r requirements.txt; fi
     - if [[ -n "${ASTROPY_VERSION}" ]]; then pip install -U "astropy${ASTROPY_VERSION}"; fi
--- a/py/desiutil/redirect.py
+++ b/py/desiutil/redirect.py
@@ -40,12 +40,12 @@ def get_libc():
             # Linux systems
             _c_stdout = ctypes.c_void_p.in_dll(_libc, 'stdout')
             _c_stderr = ctypes.c_void_p.in_dll(_libc, 'stderr')
-        except:
+        except ValueError:
             try:
                 # Darwin
                 _c_stdout = ctypes.c_void_p.in_dll(_libc, '__stdoutp')
                 _c_stderr = ctypes.c_void_p.in_dll(_libc, '__stdoutp')
-            except:
+            except ValueError:
                 # Neither!
                 pass
     return (_libc, _c_stdout, _c_stderr)
--- a/py/desiutil/redirect.py
+++ b/py/desiutil/redirect.py
@@ -212,8 +212,6 @@ def stdouterr_redirected(to=None, comm=None):
 
     _close_redirect(file)
 
-    print("redirect DONE", flush=True)
-
     if comm is not None:
         # Concatenate per-process files if we have multiple processes.
         comm.barrier()
--- a/py/desiutil/redirect.py
+++ b/py/desiutil/redirect.py
@@ -25,21 +25,30 @@ from .log import get_logger, _desiutil_log_root
 # C file descriptors for stderr and stdout, used in redirection
 # context manager.
 
-libc = ctypes.CDLL(None)
-c_stdout = None
-c_stderr = None
-try:
-    # Linux systems
-    c_stdout = ctypes.c_void_p.in_dll(libc, 'stdout')
-    c_stderr = ctypes.c_void_p.in_dll(libc, 'stderr')
-except:
-    try:
-        # Darwin
-        c_stdout = ctypes.c_void_p.in_dll(libc, '__stdoutp')
-        c_stderr = ctypes.c_void_p.in_dll(libc, '__stdoutp')
-    except:
-        # Neither!
-        pass
+_libc = None
+_c_stdout = None
+_c_stderr = None
+
+def get_libc():
+    """Helper function to import libc once."""
+    global _libc
+    global _c_stdout
+    global _c_stderr
+    if _libc is None:
+        _libc = ctypes.CDLL(None)
+        try:
+            # Linux systems
+            _c_stdout = ctypes.c_void_p.in_dll(_libc, 'stdout')
+            _c_stderr = ctypes.c_void_p.in_dll(_libc, 'stderr')
+        except:
+            try:
+                # Darwin
+                _c_stdout = ctypes.c_void_p.in_dll(_libc, '__stdoutp')
+                _c_stderr = ctypes.c_void_p.in_dll(_libc, '__stdoutp')
+            except:
+                # Neither!
+                pass
+    return (_libc, _c_stdout, _c_stderr)
 
 @contextmanager
 def stdouterr_redirected(to=None, comm=None):
@@ -62,6 +71,8 @@ def stdouterr_redirected(to=None, comm=None):
         comm (mpi4py.MPI.Comm): The optional MPI communicator.
 
     """
+    libc, c_stdout, c_stderr = get_libc()
+
     nproc = 1
     rank = 0
     MPI = None
@@ -158,7 +169,7 @@ def stdouterr_redirected(to=None, comm=None):
 
     if rank == 0:
         log = get_logger()
-        log.info("Begin log redirection to {} at {}".format(to, time.asctime()))
+        log.info("Begin log redirection to %s at %s", to, time.asctime())
 
     # Try to open the redirected file.
 
@@ -173,7 +184,7 @@ def stdouterr_redirected(to=None, comm=None):
     except:
         log = get_logger()
         log.error(
-            "Failed to open redirection file %s at %s", pto, time.asctime())
+            "Failed to open redirection file %s at %s", pto, time.asctime()
         )
         fail_open = 1
 
@@ -185,7 +196,7 @@ def stdouterr_redirected(to=None, comm=None):
         if rank == 0:
             log = get_logger()
             log.error(
-                "Failed to start redirect to %s at %s", to, time.asctime())
+                "Failed to start redirect to %s at %s", to, time.asctime()
             )
 
         _close_redirect(file)
@@ -229,7 +240,7 @@ def stdouterr_redirected(to=None, comm=None):
 
     if rank == 0:
         log = get_logger()
-        log.info("End log redirection to %s at %s", to, time.asctime()))
+        log.info("End log redirection to %s at %s", to, time.asctime())
 
     # flush python handles for good measure
     sys.stdout.flush()
--- a/py/desiutil/redirect.py
+++ b/py/desiutil/redirect.py
@@ -173,7 +173,7 @@ def stdouterr_redirected(to=None, comm=None):
     except:
         log = get_logger()
         log.error(
-            "Failed to open redirection file {} at {}".format(pto, time.asctime())
+            "Failed to open redirection file %s at %s", pto, time.asctime())
         )
         fail_open = 1
 
@@ -185,7 +185,7 @@ def stdouterr_redirected(to=None, comm=None):
         if rank == 0:
             log = get_logger()
             log.error(
-                "Failed to start redirect to {} at {}".format(to, time.asctime())
+                "Failed to start redirect to %s at %s", to, time.asctime())
             )
 
         _close_redirect(file)
@@ -229,7 +229,7 @@ def stdouterr_redirected(to=None, comm=None):
 
     if rank == 0:
         log = get_logger()
-        log.info("End log redirection to {} at {}".format(to, time.asctime()))
+        log.info("End log redirection to %s at %s", to, time.asctime()))
 
     # flush python handles for good measure
     sys.stdout.flush()
--- a/py/desiutil/redirect.py
+++ b/py/desiutil/redirect.py
@@ -1,9 +1,9 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 # -*- coding: utf-8 -*-
 """
-===================
+=================
 desiutil.redirect
-===================
+=================
 
 Utilities for redirecting stdout / stderr to files.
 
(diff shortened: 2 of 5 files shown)


tskisner:
The MPI test script cannot be spawned from within the unit test runner, so I simply print out the command to run.

araichoor:
I've run two checks (no MPI).
One on a small piece of code which produces an error, and one on a typical call of fba_cmx, with an (intentional) error.
Both work fine, the error message is printed in the log file (and also on the prompt).

tskisner:
There is one transient test failure unrelated to this PR, but the rest seems to be working.  After this work is merged we can use this in fiberassign and switch desispec over to this version of the redirection.

.travis.yml
```diff
@@ -95,7 +95,7 @@ install:
     - pip install -U pip
     - if [[ "${MAIN_CMD}" == pycodestyle ]]; then pip install pycodestyle; fi
     - if [[ "${MAIN_CMD}" == sphinx-build* ]]; then pip install Sphinx; fi
-    - if [[ "${MAIN_CMD}" == pytest ]]; then pip install -r requirements.txt; fi
+    - if [[ "${MAIN_CMD}" == pytest ]]; then pip install -r requirements.txt; pip install . ; fi
```
weaverba137, line 98: This seems like a very sneaky change.  Could you please explain in detail why this is necessary?  Test were working perfectly fine without this prior to adding just one module.

py/desiutil/redirect.py
```diff
@@ -0,0 +1,247 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+"""
+===================
```
weaverba137, line 4: Please make sure that this doc string is valid RST.  I notice the `=` do not match the length of the title.

py/desiutil/redirect.py
```diff
@@ -0,0 +1,247 @@
...
+
+from contextlib import contextmanager
+
+from .log import get_logger, _desiutil_log_root
+
+
+# C file descriptors for stderr and stdout, used in redirection
+# context manager.
+
+libc = ctypes.CDLL(None)
```
weaverba137, line 28: Why does this code need to be called at the module level?  Calling functions at import time has been the source of much pain in the past.

py/desiutil/redirect.py
```diff
@@ -0,0 +1,247 @@
...
+        pto = "{}_{}".format(to, rank)
+
+    fail_open = 0
+    file = None
+    try:
+        file = _open_redirect(pto)
+    except:
+        log = get_logger()
+        log.error(
+            "Failed to open redirection file {} at {}".format(pto, time.asctime())
```
weaverba137, line 176: The logging framework actually prefers `%` based formatting, *e.g.*:
```
python
log.error("Failed to open redirection file %s at %s.", pto, time.asctime())
```
This style is used through most of the rest of desiutil.  It also allows the logging framework to perform certain optimizations.

py/desiutil/redirect.py
```diff
@@ -0,0 +1,247 @@
...
+    # Output should now be redirected.  Run the code.
+
+    fail_run = 0
+    try:
+        yield # Allow code to be run with the redirected output
+    except:
+        # We have an unhandled exception.  Print a stack trace to the log.
+        exc_type, exc_value, exc_traceback = sys.exc_info()
+        lines = traceback.format_exception(exc_type, exc_value, exc_traceback)
+        print("".join(lines), flush=True)
```
weaverba137, line 206: This bare `print()` is safe?

py/desiutil/test/test_redirect.py
```diff
@@ -0,0 +1,148 @@
...
+        generate_output()
+
+"""
+
+
+class TestRedirect(unittest.TestCase):
+    """Test desiutil.redirect"""
+
+    @classmethod
+    def setUpClass(cls):
```
weaverba137, line 74: Add `tearDownClass(cls)` to clean up this temporary directory.

weaverba137 (changes requested):
This is a very complex module, and I will probably need to take more time than this to fully digest it.  So this *first* review is just to catch the obvious things.  I'll do a second review after the first round is addressed.

.travis.yml
tskisner: Previously, the unit tests were running "in place" from setup.py.  This change installs the package first before running the tests.  Many packages do this in order to test that the installation process produces a working, installed package.  Example:

https://github.com/desihub/fiberassign/blob/master/.travis.yml#L135

In this particular instance, I am generating and running external python scripts to test the redirection, because this redirection does not work inside the python unittest framework (the unittest runner has a handle to stdout during the tests, and redirecting it invalidates that handle, raising an I/O exception).

In order to have an external script run the redirection, desiutil must be installed and importable.  That is the reason behind this change.

weaverba137: The problem with that in the future we will not be using `setup.py` for testing anymore.  In fact, desiutil is already using `pytest` instead, and we will probably eventually move to `tox`.  So I don't think this change is future-proof.

py/desiutil/redirect.py
tskisner: This occurs while the stdout / stderr is being redirected to a file, and the wrapped code block just raised an exception.  So the stack trace will end up in the redirected file.  I think it is fine...

py/desiutil/test/test_redirect.py
tskisner: Done

py/desiutil/redirect.py
tskisner: Done

py/desiutil/redirect.py
tskisner: Fixed

py/desiutil/redirect.py
tskisner: That block of code gets a handle to the system `libc` and the low-level `STDOUT` and `STDERR`.  We definitely only want to do that once, so it seemed fine to do that on module import.  Of course I can make a function that sets those module level variables if they are not set and returns them.  However, for the record this code is copied out of desispec where it has been working well for years.

tskisner:
I'll go wrap the libc handle fetching in a function.  Just for the record, although this is "complicated" code, it is a minor reworking of code that has been used in `desispec` for a long time:

https://github.com/desihub/desispec/blob/master/py/desispec/parallel.py#L292

This version of the code has been improved to be more resilient against the wrapped code raising an exception.  And this version of the code now has explicit unit tests, unlike the desispec version.

py/desiutil/redirect.py
tskisner: This is now wrapped in a function

.travis.yml
tskisner: If we were using a tool like `tox`, then the package would be automatically installed before running the tests (see https://tox.readthedocs.io/en/latest/#current-features), and so this workaround would not be needed.  It is only needed because we are running the tests using `setup.py`.

Let me know if there are any other changes needed.  I think I have resolved all your previous review points.  The fiberassign commissioning script is ready to use this function for its logging once it is merged.

weaverba137:
OK, but again, we are *not* using setup.py for tests now.  `pytest` is used instead.  I am pretty sure that the travis tests already reflect this.  Where are the remaining tests being run that still use `python setup.py test`?

.travis.yml
weaverba137: OK, but again, we are not using setup.py for tests now. pytest is used instead. I am pretty sure that the travis tests already reflect this. Where are the remaining tests being run that still use python setup.py test?

tskisner: Sure.  pytest best practices (https://docs.pytest.org/en/latest/goodpractices.html) recommend installing the package in editable mode first.  In our case we are running tests once in a CI instance, so installing in normal mode is fine.  I'm confused what changes you are looking for here?  

From my perspective, testing against the installed package has the benefit of testing the install process as well.  And as previously noted, for this type of redirection there is no other way to test it.

weaverba137: Maybe we're talking past each other here, and should move this to an actual audio conversation.  But let me try one more time to make myself clear.

* Travis tests were already working *with* `pytest` and *without* this change.
* You have not shown me any cases where `python setup.py test` is still being used.
* Therefore, I do not understand why this change is necessary.  Best practice or no best practice.
* This PR is supposed to be about this redirect module, not changing the nature of the Travis tests.

tskisner: 1. `pytest` is currently being used *without* installing the package first.  The earlier statement about `setup.py` was an error.  The point is that the package is not being installed first before running the tests.

2.  Quoting my earlier statement in this same thread:  _"I am generating and running external python scripts to test the redirection, because this redirection does not work inside the python unittest framework ...  In order to have an external script run the redirection, desiutil must be installed and importable. That is the reason behind this change."_

3.  I am only changing the travis file to install the package first, as required in order to test the redirection function.

4.  My statements / links about "best practices" were simply to counter your seeming resistance to installing the package prior to testing.  I was just trying to communicate that such a step is a perfectly reasonable thing to do and is already done automatically with some test runners like `tox`.

weaverba137: > desiutil must be installed and importable

Then why not set PYTHONPATH?

weaverba137:
@sbailey, I'd like your review here.  We're getting bogged down in changes to Travis tests that *I* feel are not necessary, but we have not come to a consensus.  I also don't know how urgent this is, and because of the test issue, I have not had time to review the module really thoroughly for other problems.  Could you please chime in?

tskisner:
@weaverba137 , do you understand at least why I need to install the package first?  I really am not trying to make more work for myself just for fun.  The python `unittest` module keeps a persistent handle to stdout.  It is not possible to redirect stdout from inside a unittest.  Thus the need to run an external script.

Just to reiterate- a messier version of this code (with no unit tests at all) is **already in use** in desispec.  This PR was an effort to put the code in a common place and clean it up and make some tests.

weaverba137:
@tskisner, honestly, no, I do not understand that, in part because I haven't had time to dig into the details of the module itself.  Without delving into that, I don't know whether there is a more clever way to manipulate `unittest` to do what we need.

sbailey:
Wearing my manager hat, summarizing my understanding of this PR:
  * @tskisner is adding code that has been in use in desispec without problems for a couple of years.
  * Moving it to desiutil allows it to be used by other desi products without introducing dependency loops or new desi package cross dependencies (good!)
  * At the same time, he is adding unit tests (good!); it was not previously unit-tested in desispec.
  * The unittest module keeps a persistent handle to stdout, which conflicts with the new functionality of redirecting stdout, so the new tests spawn a script to work around this while still testing the new functionality.  However, that test script needs desiutil to be installed so that it can `import desiutil.redirect`.
    * For reasons I don't completely understand, "python setup.py test" does a more complete test installation environment than "pytest" does, such that if desiutil is not installed prior to testing, the tests in this branch work with "python setup.py test" but not "pytest".
    * however, testing via setup.py / setuptools is deprecated by the Python gods (not by us), thus the motivation to move to pytest or other.
  * I don't think it was explicitly stated here, but my memory of previous conversations is that the motivation for being able to test a package without installing it is so that you can catch any errors specific to your local system prior to installing potentially broken code.  So the objection to the new "install before testing" requirement isn't just "but we haven't needed that before so why do we need it now", but rather that there are specific motivations for being able to test first.
  * At the same time, purity of testing is getting in the way of actually having useful code put in place where it can be used, and including automated tests on it.
  * And having Travis install before testing checks the installation process itself, which in other packages has occasionally been problematic, so that can be viewed as a feature not a bug, even if it is a step backwards on testing a different case.

Executive decision:
  * Keep the install before testing change in this PR for Travis so that this new code can be included in the automated testing
  * Add an import check to the test so that it doesn't break the ability to test the rest of desiutil prior to installation (I just added that, tested at NERSC on a git clone of this branch without desiutil installed).
    * Note that if you are using "pytest" to test prior to installing on some system, it skips testing this new code, but that is no worse than the previous state which didn't test it at all anywhere, and at least for awhile longer "python setup.py test" does the full suite of tests including `desiutil.redirect` even if desiutil isn't installed.
    * Caveat: if you have a working installed copy of desiutil, and then have a new branch that has broken `desiutil.redirect` but that somehow got through Travis, and you are trying to test that prior to replacing your current installation, I'm not entirely sure what would happen.  However, that seems like such a theoretical corner case that I don't think it warrants blocking the other benefits of getting this code in here with automated travis testing.

weaverba137:
@sbailey, thank you.  I also remember now that we had issues in the past along the lines of "is this testing the bare git clone, or code that is installed somewhere else".

I will take one more look at the module itself this afternoon, and I will approve and merge after that.

I added `desiutil.redirect` to the API documentation and updated the change log.

weaverba137:
Is the function `get_libc()` intended to be part of the public API, or is it used only internally by the redirect module?  If the latter, I'd recommend calling it `_get_libc()`.

weaverba137:
```python
                _c_stdout = ctypes.c_void_p.in_dll(_libc, '__stdoutp')
                _c_stderr = ctypes.c_void_p.in_dll(_libc, '__stdoutp')
```
Is this really the correct calling sequence for MacOS/Darwin?  Stdout and stderr have exactly the same names?

weaverba137:
```python
        if sys.version_info[0] < 3:
            sys.stdout = os.fdopen(fd_out, "wb")
            sys.stderr = os.fdopen(fd_err, "wb")
        else:
            # Python 3 case
            sys.stdout = io.TextIOWrapper(os.fdopen(fd_out, 'wb'))
            sys.stderr = io.TextIOWrapper(os.fdopen(fd_err, 'wb'))
```
I'm pretty sure we don't need to support Python 2 in this module.

weaverba137:
```python
    def _open_redirect(filename):
        # Open python file, which creates low-level POSIX file
        # descriptor.
        file_handle = open(filename, "w")
        # ...
```
In most other places, files are opened `"wb"` (binary write).  Is this file handle deliberately meant to be opened `"w"` (text write)?

weaverba137:
```python
log.info("Begin log redirection to %s at %s", to, time.asctime())
```
Since you're using `desiutil.log`, a timestamp is already provided by the log.  Is it important to log two separate timestamps?

weaverba137:
Why are there so many calls to `log = get_logger()`?  You only need one.

weaverba137:
Now that I've finally looked over the module in detail, there are several more points to address before this can be merged.  Hopefully these can be done quickly.

tskisner:
I made these changes:

- fix typo stdoutp --> stderrp on Darwin (I can't test this, but it sure looks like a typo)
- remove timestamps in log messages
- remove python2 check
- open files in wb mode

I am intentionally getting a new log handle each time I need it, since I recall problems with this in past.  Likely something to do with the redirect modifying the logger.

weaverba137 (approved):

weaverba137:
I verified that `ctypes.c_void_p.in_dll(_libc, "__stderrp")` does give a valid result on MacOS/Darwin.  Merge when ready.

tskisner:
Thanks @sbailey .  Looks like travis may never run, so merge if you are happy with it.

sbailey:
Two updates:
  * fixed `_get_libc` vs. `get_libc` typo (missing `_` when using it)
  * logger only prints a timestamp if specifically requested via `get_logger(timestamp=True)`, so I added that to the `get_logger` calls for beginning and ending the redirect which are where we most want the timing statements.

Travis is completely bogged down (tests aren't getting through on multiple repos), but I verified that they pass on my laptop (Mac) and at NERSC so I'll merge anyway without Travis so that we can proceed with a new tag.  This will flag that PR as failing tests, but ok.

PR #156, opened 2020-12-10
GitHub actions

### LINKED ISSUE #154: Don't assume master is default branch
In `desiInstall` and other functionality, allow the default git branch to be something other than `master`.

@sbailey, I believe the GitHub default for new repositories is to call the default branch `main`.  I think we should follow this convention.

weaverba137:
This PR

* Fixes #154
  - The cost of eliminating `master` is that all branches must be explicitly labeled as such.  For example `desiInstall desiutil branches/main`.  *Exception:* `desiInstall foo trunk` is an abbreviation for `desiInstall foo branches/trunk`.
  - This may require updates to *e.g.* integration tests that install (checkout/clone) from scratch into a clean environment.  However, I think the desispec nightly integration test doesn't actually do that, it just updates existing checkouts, right?
* Also implements GitHub Actions for desiutil.
  - However, tests of desiutil against Astropy 2.0.x are abandoned because it is not worth finding a compatible version of healpy that compiles in the GitHub environment.


sbailey (approved):
This looks great.  The GitHub Actions are a breath of fresh air.  I tested desiInstall for a tag and a branch and both worked.  Minor items for consideration but not blocking factors for merging:
* Consider shortening the names of the tests in `.github/workflows/python-package.yml`.  When viewing the summary of tests in the github PR interface, the right side is getting cutoff (e.g. "CI / Check code style; allow failures. (ubuntu-latest, 3.8) (pull_request) Suc") even when I have the browser window maximally wide on my laptop, and the interface doesn't provide a horizontal slider.  I presume that if the test failed it would be flagged obviously, but short names would help, e.g. something like "test", "coverage", "docs", "style".
* very minor: could "desiInstall (package) main" be a shortcut for "desiInstall install branches/main" similar to "desiInstall (package) trunk"?

weaverba137:
@sbailey, good suggestions.  I'll get those in before merging.

weaverba137:
@sbailey, do you have other significant changes to desiutil planned for release 20.12?  If not, I can tag desiutil after this merge.

sbailey:
No other planned changes for desiutil for 20.12; please tag after merging.  Thanks.

PR #159, opened 2020-12-17
Support recent astropy versions

weaverba137:
[Astropy 4.2 no longer has `astropy.erfa`](https://docs.astropy.org/en/stable/whatsnew/4.2.html#whatsnew-4-2-erfa), so ERFA warnings need to support the new package.

In addition, some unexpected interaction between very recent versions of `setuptools` and `disutils` were discovered, as more recent versions of Astropy bring in these more recent versions of `setuptools`.


sbailey (approved):
Looks good, thanks.

PR #160, opened 2021-02-03
allow for cmake installation

marcelo-alvarez:
This modification changes desiInstall such that if there is a `CMakeLists.txt` file in the top-level directory of a package, then it will use the commands `cmake .; make -j 8` to install. This currently is for specex but should work generally and can be modified as necessary in the future.

### DIFF
--- a/py/desiutil/install.py
+++ b/py/desiutil/install.py
@@ -516,12 +516,16 @@ class DesiInstall(object):
             self.log.debug("Forcing build type: make")
             build_type.add('make')
         else:
+            print('cwd: ',self.working_dir);
             if os.path.exists(os.path.join(self.working_dir, 'setup.py')):
                 self.log.debug("Detected build type: py")
                 build_type.add('py')
             if os.path.exists(os.path.join(self.working_dir, 'Makefile')):
                 self.log.debug("Detected build type: make")
                 build_type.add('make')
+            elif os.path.exists(os.path.join(self.working_dir,'CMakeLists.txt')):
+                self.log.debug("Detected build type: cmake")
+                build_type.add('cmake')
             else:
                 if os.path.isdir(os.path.join(self.working_dir, 'src')):
                     self.log.debug("Detected build type: src")
@@ -834,11 +838,14 @@ class DesiInstall(object):
             # installation or we still need to compile the C/C++ product
             # (we had to construct doc/Makefile first).
             #
-            if 'make' in self.build_type or 'src' in self.build_type:
+            if ('make' in self.build_type or 'src' in self.build_type or
+                'cmake' in self.build_type):
                 if 'src' in self.build_type:
                     command = ['make', '-C', 'src', 'all']
-                else:
+                elif 'make' in self.build_type:
                     command = ['make', '-j', '8', 'install']
+                elif 'cmake' in self.build_type:
+                    command = ['cmake .; make -j 8']
                 self.log.debug(' '.join(command))
                 if self.options.test:
                     self.log.debug("Test Mode.  Skipping 'make install'.")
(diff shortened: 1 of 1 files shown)


sbailey:
@weaverba137 this PR appears to have trigged a bunch of MaskedArrayWithLimits-related test failures completely unrelated to the changes in this PR.  Please investigate.

py/desiutil/install.py
```diff
@@ -516,12 +516,16 @@ def build_type(self):
             self.log.debug("Forcing build type: make")
             build_type.add('make')
         else:
+            print('cwd: ',self.working_dir);
```
sbailey, line 519: Please remove bare `print` or switch to useing `self.log.debug`.

py/desiutil/install.py
```diff
@@ -847,7 +854,11 @@ def install(self):
                         os.chdir(self.install_dir)
                     else:
                         os.chdir(self.working_dir)
-                    proc = Popen(command, universal_newlines=True,
+                    if 'cmake' in self.build_type:                        
+                        proc = Popen(command, universal_newlines=True,
+                                 stdout=PIPE, stderr=PIPE, shell=True)
```
sbailey, line 859: This seems messy / fragile where "cmake" requires two-commands-in-one and thus a string command with `shell=True`, while "make" uses a single-command-passed-as-a-list with default `shell=False`.  The normal admonishment against `shell=True` [[link](https://docs.python.org/3/library/subprocess.html#security-considerations)] is due to security concerns about shell injection vulnerabilities from unsanitized user input strings, but in this case we have completely control over the commands created without any user-provided options, and thus I think both "make" and "cmake" could use string (not list) commands with `shell=True` to avoid this if/else.  Thoughts?

sbailey (changes requested):
Looks ok, with minor cleanup requests.

Let's get unreleated test failures sorted out first, perhaps in a separate PR and then re-merged into this one.

py/desiutil/install.py
weaverba137: To further @sbailey's comment, did you actually *test* the commands with `shell=False` and come to the conclusion that it was impossible to run with `shell=False`?

I don't think anyone would object if we ran two commands, both with `shell=False`, if the one and only reason for `shell=True` is a `;`.

tskisner:
Just throwing out another option- it is possible to have a `setup.py` which calls CMake to do the actual building of extensions.  See for example:  https://github.com/hpc4cmb/toast/blob/master/setup.py

This even supports setting environment variables to pass options to the underlying cmake call.

This way the desiInstall tools can just run setup.py and not worry about what it is doing under the hood.  It is just a question of where you want to put the extra logic- in the package or in the external tools.

py/desiutil/install.py
marcelo-alvarez: Did a test and found it would require two running two commands and the `shell=True` was chosen. Will try the `setup.py` solution @tskisner mentioned below and if so these changes will be unnecessary.

marcelo-alvarez:
> Just throwing out another option- it is possible to have a `setup.py` which calls CMake to do the actual building of extensions. See for example: https://github.com/hpc4cmb/toast/blob/master/setup.py
> 
> This even supports setting environment variables to pass options to the underlying cmake call.
> 
> This way the desiInstall tools can just run setup.py and not worry about what it is doing under the hood. It is just a question of where you want to put the extra logic- in the package or in the external tools.

I think having cmake called by `setup.py` is preferable and will look into it, thanks for pointing this out.

py/desiutil/install.py
weaverba137: That sounds promising.  I was about to point out that you have `if 'cmake' in self.build_type:`, but you have not modified the code in any way to *detect* `'cmake'` as the `build_type`.

weaverba137:
Requiring `numpy<1.20` fixed the masked array problems reported.  I'll spin that to a separate issue.

marcelo-alvarez:
The specex build now uses python setup.py instead of cmake, so this change is no longer necessary. Closing this pull request and deleting the branch.

PR #163, opened 2021-02-11
install py or src not both; no silent installdir fallback

sbailey:
Our two hybrid python/C++ packages (fiberassign and specex) have now converted to using `python setup.py install` for installations, including compiling the C++ code.  This was creating a fake-failure problem for desiInstall which was noticing the `src/` directory and trying to run a non-existent Makefile.  This PR updates the installation code to use build type "py" OR "make" OR "src" but not "py" AND "src".  I left the "make" and "src" logic in place even though we don't have any packages using them now, but if the `setup.py` file is found it uses that and then moves on.

This also fixes a bug where if the `desiInstall --root ...`directory didn't exist at NERSC, the code was silently falling back to the default NERSC install location instead of warning the user that their requested installation location doesn't exist and stop.

With current master:
```
[cori12 temp] desiInstall -r $SCRATCH/desi/temp fiberassign 2.1.1
CRITICAL:install.py:866:install:2021-02-11T15:15:02: Error during compile: make: *** src: No such file or directory.  Stop.
```
That produces a useable installation, albeit without the permissions fully corrected.

With this branch there are no errors, fully successful install.

### DIFF
--- a/doc/desiInstall.rst
+++ b/doc/desiInstall.rst
@@ -207,11 +207,9 @@ Determine Build Type
 --------------------
 
 The downloaded code is scanned to determine the build type.  There are several
-possible build types that are *not* mutually exclusive.
+possible build types that are mutually exclusive.  They are derived in this
+order and the first matching method is used:
 
-plain
-    This is the default build type.  With this build type, the downloaded code
-    is simply copied to the final install directory.
 py
     If a setup.py file is detected, :command:`desiInstall` will attempt to execute
     :command:`python setup.py install`.  This build type can be suppressed with the
@@ -224,6 +222,9 @@ src
     :command:`desiInstall` will attempt to execute :command:`make -C src all`.  This build type
     *is* mutually exclusive with 'make', but is not mutually exclusive with
     the other types.
+plain
+    If no other build type is detected, the downloaded code
+    is simply copied to the final install directory.
 
 **It is the responsibility of the code developer to ensure that these
 build types do not conflict with each other.**
--- a/doc/desiInstall.rst
+++ b/doc/desiInstall.rst
@@ -219,15 +219,13 @@ make
     :command:`make install`.
 src
     If a Makefile is not present, but a src/ directory is,
-    :command:`desiInstall` will attempt to execute :command:`make -C src all`.  This build type
-    *is* mutually exclusive with 'make', but is not mutually exclusive with
-    the other types.
+    :command:`desiInstall` will attempt to execute :command:`make -C src all`.
 plain
     If no other build type is detected, the downloaded code
     is simply copied to the final install directory.
 
-**It is the responsibility of the code developer to ensure that these
-build types do not conflict with each other.**
+**It is the responsibility of the code developer to understand these build
+types and choose the one appropriate for the package being developed.**
 
 Determine Install Directory
 ---------------------------
--- a/doc/desiInstall.rst
+++ b/doc/desiInstall.rst
@@ -104,7 +104,7 @@ Directory Structure Assumed by the Install
 :command:`desiInstall` is primarily intended to run in a production environment that
 supports Module files, *i.e.* at NERSC.
 
-*:command:`desiInstall` does not install a Modules infrastructure for you.* You have to
+:command:`desiInstall` *does not install a Modules infrastructure for you.* You have to
 do this yourself, if your system does not already have this.
 
 For the purposes of this section, we define ``$product_root`` as the
@@ -145,7 +145,7 @@ etc/
     Miscellaneous metadata and configuration.  In most packages this only
     contains a template Module file.
 lib/pythonX.Y/site-packages/
-    Contains installed Python code.  ``X.Y`` would be ``2.7`` or ``3.5``.
+    Contains installed Python code.  ``X.Y`` would be, *e.g.*, ``3.6`` or ``3.8``.
 py/
     Sometimes we need to install a git checkout rather than an installed package.
     If so, the Python code will live in *this* directory not the ``lib/``
@@ -253,7 +253,7 @@ Module Infrastructure
 
 :command:`desiInstall` sets up the Modules infrastructure by running code in
 :mod:`desiutil.modules` that is *based on* the Python init file supplied by
-the Modules infrastructure, but updated to be both Python 2 and Python 3 compatible.
+the Modules infrastructure.
 
 Find Module File
 ----------------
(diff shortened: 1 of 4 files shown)


weaverba137:
I think this is fine at first glance, but there is also the desiInstall.rst documentation that will need to be changed.  Could you do that?

weaverba137:
Even though these are vestigial, is the remaining logic "make" XOR "src"?  That is, all possible build types are mutually exclusive?

sbailey:
My intension was to make all build types exclusive (make and src already were, but previously they could be combined with py).  I'll update docs.

weaverba137:
Good, thank you.  I'll do a formal review soon, but it might have to wait until tomorrow morning.

doc/desiInstall.rst
```diff
@@ -224,6 +222,9 @@ src
     :command:`desiInstall` will attempt to execute :command:`make -C src all`.  This build type
     *is* mutually exclusive with 'make', but is not mutually exclusive with
```
weaverba137, line 223: This contradicts the statement above that the build types are mutually exclusive.

doc/desiInstall.rst
```diff
@@ -224,6 +222,9 @@ src
...
     the other types.
+plain
+    If no other build type is detected, the downloaded code
+    is simply copied to the final install directory.
 
 **It is the responsibility of the code developer to ensure that these
```
weaverba137, line 229: Since the build types are now mutually exclusive, perhaps we should rephrase: "It is the responsibility of the code developer to understand these build types and choose the one appropriate for the package being developed."

py/desiutil/install.py
```diff
@@ -573,17 +573,16 @@ def set_install_dir(self):
...
-        except KeyError:
-            self.nersc = None
-        if self.options.root is None or not os.path.isdir(self.options.root):
-            if self.nersc is not None:
-                self.options.root = self.default_nersc_dir()
-            else:
-                message = "Root install directory is missing or not set."
-                self.log.critical(message)
-                raise DesiInstallException(message)
+        self.nersc = os.getenv('NERSC_HOST') #- will be None if not set
```
weaverba137, line 576: Is this particular change really necessary?  I really hate `os.getenv()` because it never throws an exception, so it's much harder to tell when a variable is undefined.

py/desiutil/install.py
```diff
@@ -573,17 +573,16 @@ def set_install_dir(self):
...
+        if self.options.root is None and self.nersc is not None:
+            self.options.root = self.default_nersc_dir()
+
+        if (self.options.root is None or not os.path.isdir(self.options.root)) \
```
weaverba137, line 580: Use parenthesis to break across lines.  The backslash is ancient history.

py/desiutil/test/test_install.py
```diff
@@ -326,7 +326,7 @@ def test_set_install_dir(self):
         with patch.dict('os.environ', {'NERSC_HOST': 'edison', 'DESI_PRODUCT_ROOT': 'FAKE'}):
             del environ['DESI_PRODUCT_ROOT']
             test_code_version = 'test-blat-foo'
-            options = self.desiInstall.get_options(['desiutil', test_code_version])
+            options = self.desiInstall.get_options(['desiutil', test_code_version, '--test'])
```
weaverba137, line 329: Could you explain in more detail why this change was necessary?

weaverba137 (approved):
I found a few minor issues to discuss.

weaverba137 (changes requested):
Please see my other review.  This is just to ensure that a "Request changes" result is returned.

doc/desiInstall.rst
sbailey: Thanks for catching that; I corrected it.

doc/desiInstall.rst
sbailey: Good; rephrased with your suggestion.

py/desiutil/install.py
sbailey: OK.  I reverted the `os.getenv()` to use `try/except` around `os.environ` again, while noting that simply catching the exception and setting the variable to `None` is exactly the same thing that `os.getenv` does if the environment variable isn't defined.  I agree that `os.getenv()` is a landmine if the code assumes that the variable is defined and would choke if silently given a `None`, but in this case `None` is the correct fallback, so `os.getenv` seemed simpler.  (And now I've spent 7 lines of text explaining why I prefered 1 line of code over 4... :)

I think you were only commenting on the `os.getenv()` line, but just to be clear: the rest of the changes here are important, because they are fixing the problem where desiInstall was silently falling back to the default NERSC location if the user-specified root directory didn't exist.

py/desiutil/install.py
sbailey: I have a gray beard and have been using python since before numpy existed.   But ok, parentheses it is...

py/desiutil/test/test_install.py
sbailey: `set_install_dir()` will now raise an exception if the installation directory doesn't exist, which was tripping up the test that was using the fallback to the NERSC default location, which doesn't exist in the test environment.  When actually installing at NERSC, I think it should be an error if the default location doesn't exist, so I used the `--test` hooks to prevent the existence check exception in `set_install_dir` while testing.

sbailey:
@weaverba137 thanks for the comments / suggestions.  I think I have addressed all of them.  Please re-review.

py/desiutil/install.py
weaverba137: I realize both old and new have the same result. So why change it?

Cosmetic changes like this distract from the real changes. I've heard that before in some of my own pull requests.

Although there apparently is no active unit test on this bit of code, the try: clause would potentially be something that is deliberately triggered in a unit test.

py/desiutil/test/test_install.py
weaverba137: OK.  I'll take a look and see if the unit test can be made a bit more robust.

py/desiutil/install.py
sbailey: Noted and agreed about mixing cosmetic changes distracting from the substantive changes.  In this case it was the result of several iterations of sorting out what this code should do and various attempts to find the best order of if/then try/except for what should be checked in what order, and it landed back there in the N-1 iteration.  I have reverted it back your preferred method.

weaverba137 (approved):

weaverba137:
OK to merge now.  I need to do an overhaul of the desiutil testing infrastructure anyway but I will punt that to a separate issue.  I can create a tag after merge if you're ready for that.

sbailey:
Thanks, merging now.  I may have another PR or two for desiutil today prior to a tag for 21.2 / Cascades, so no tag yet.

PR #166, opened 2021-03-15
add a function that converts SFD EBV to 'true' EBV

segasai:
Hi, 

To address issues raised here 
https://github.com/desihub/desispec/pull/1159

I've added a function that converts SFD EBV to 'true' EBV, in order to avoid having a 0.86 normalization number in multiple places.

(To avoid possible  problems related to diving by .86 instead of multiplying, I explicitly take input EBVs rather than just return the normalization number.)

### DIFF
--- a/py/desiutil/dust.py
+++ b/py/desiutil/dust.py
@@ -82,6 +82,19 @@ def mwdust_transmission(ebv, band, photsys):
 
         return transmission
 
+def convert_sfd_ebv_to_true_ebv(ebv):
+    """
+    Apply the correction from Schlafly+2010
+    https://ui.adsabs.harvard.edu/abs/2010ApJ...725.1175S/abstract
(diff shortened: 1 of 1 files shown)


sbailey:
After the 0.1% run is over (next week?) @julienguy and @schlafly will standardize our `desitutil.dust` calculations and merge this 0.86 factor internally so that the user doesn't have to do it externally (which would likely result in some users not knowing to do so).  Let's not merge this PR yet; they can decide if they want to build on on this PR or do it separately.

I do like the spirit of `corrected_ebv = function(sfd_ebv)` to avoid multiply vs. divide confusion.

weaverba137:
This PR is superseded by #169.

julienguy:
Closing because superseded by more recent PR.

PR #168, opened 2021-03-29
Switch to pip install

### LINKED ISSUE #167: Migrate desiInstall to pip
Starting with recent versions of Python (`3.[89]`), `python setup.py install` is not producing installs that are importable, *e.g.* `import desiutil` fails.  `pip install --prefix ...` does appear to work.  desiInstall should:

* Migrate to pip
* Use `pip install git+https://github.com/desihub/desiutil.git...` instead of needing to download and open a tarball.

weaverba137:
This PR closes #167.  In addition to a simple replacement of `python setup.py install` with the equivalent `pip install .`, this PR

* Updates the known packages.
* Begins preparation for the transition to Lmod-based modules.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -2,10 +2,12 @@
 Change Log
 ==========
 
-3.1.3 (unreleased)
+3.2.0 (unreleased)
 ------------------
 
-* No changes yet
+* Use ``pip install .`` instead of ``python setup.py install`` (PR `#168`_).
+
+.. _`#168`: https://github.com/desihub/desiutil/pull/168
 
 3.1.2 (2021-02-15)
 ------------------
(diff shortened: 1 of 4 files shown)


sbailey (approved):
Looks good, thanks.

It looks like the `pip_safe` list in install.py isn't used yet.  Is that correct?  ok to leave it as is, or consider `_pip_safe` hiding it or commenting it out if not needed.  Since we have a mix of packages, I like that the current method appears to automatically detect what to do rather than requiring a proactive opt-in list to be maintained, so we may never need this.

weaverba137:
Indeed, `pip_safe` is "reserved for future use".  I'll comment it out for now, then merge & tag.

PR #176, opened 2021-08-02
Ignore messages on STDERR unless there is a non-zero return code

### LINKED ISSUE #174: desiInstall pip warnings resulting in CRITICAL level errors
When using pip 21.2.1 and python 3.9.6 in a test conda environment, pip prints a deprecation message that desiInstall interprets as a CRITICAL-level error and exists with a non-zero exit code and leaves the git directory behind.  Investigate whether we need to do anything about this warning, and update desiInstall so that it doesn't treat it as a fatal error.

Steps to reproduce at NERSC:
```
module load python/3.8-anaconda-2020.11
conda create --name testpip --yes python=3.9
source activate testpip
conda install requests numpy astropy --yes

pip install git+https://github.com/desihub/desiutil.git
export LANG=en_US.utf8
desiInstall -r $SCRATCH/temp desispec master
```

Results in an exit code 1 and the following messages:
```
WARNING:install.py:183:get_options:2021-07-26T22:19:16: The environment variable DESICONDA is not set!
CRITICAL:install.py:854:install:2021-07-26T22:19:24: Error during installation:   DEPRECATION: A future pip version will change local packages to be built in-place without first copying to a temporary directory. We recommend you use --use-feature=in-tree-build to test your packages with this new behavior before it becomes the default.
   pip 21.3 will remove support for this functionality. You can find discussion regarding this at https://github.com/pypa/pip/issues/7555.
```

weaverba137:
This PR fixes #174 by changing the logic of how messages printed on STDERR are handled by `desiInstall`.  The new logic can be summarized by this snippet:

```python
                    if len(err) > 0:
                        #
                        # Pass STDERR messages to the user, but do not
                        # raise an error unless the return code was non-zero.
                        #
                        if proc.returncode == 0:
                            message = ("Pip emitted messages on STDERR; these can probably be ignored:\n" +
                                       err)
                            self.log.warning(message)
                        else:
                            message = ("Potentially serious error detected during pip installation:\n" +
                                       err)
                            self.log.critical(message)
                            raise DesiInstallException(message)
```

@sbailey, I know you're out of the office.  Maybe @tskisner can take a quick look?  I think this will be pretty non-controversial though.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -6,8 +6,11 @@ Change Log
 ------------------
 
 * Optionally compute the MW dust transmission in the WISE bands (PR `#175`_).
+* Do not treat messages printed on STDERR as errors during :command:`desiInstall` (PR `#176`_).
 
 .. _`#175`: https://github.com/desihub/desiutil/pull/175
+.. _`#176`: https://github.com/desihub/desiutil/pull/176
+
 
 3.2.2 (2021-06-03)
 ------------------
--- a/py/desiutil/install.py
+++ b/py/desiutil/install.py
@@ -833,24 +833,16 @@ class DesiInstall(object):
                     self.log.debug(out)
                     if len(err) > 0:
                         #
-                        # Some warnings can be produced by processing
-                        # MANIFEST.in and not finding directories.  These
-                        # can be ignored.
+                        # Pass STDERR messages to the user, but do not
+                        # raise an error unless the return code was non-zero.
                         #
-                        manifestre = re.compile(r"(warning: |)no" +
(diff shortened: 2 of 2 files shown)


sbailey (approved):
Thanks, this looks good.  I tested it with the case in #174 and it works as expected (printing a warning with the pip stderr message but not generating a CRITICAL-level error with a non-zero exit code).

weaverba137:
@sbailey, do we need to get this into a tag right away or is it non-urgent?

sbailey:
non-urgent, we can tag later at our leisure.  Thanks.

PR #181, opened 2021-12-11
fix pcolormesh grid deprecation warning with matplotlib 3.5

sbailey:
This PR avoids a deprecation warning when using matplotlib 3.5:
```
pytest py/desutil/test
...
test/test_plots.py::TestPlots::test_plot_sky_binned
  /global/common/software/desi/users/sjbailey/desiutil/py/desiutil/plots.py:718: MatplotlibDeprecationWarning: Auto-removal of grids by pcolor() and pcolormesh() is deprecated since 3.5 and will be removed two minor releases later; please call grid(False) first.
    mesh = ax.pcolormesh(grid_ra, grid_dec,
```

However, I haven't looked into what is actually happening with turning the grid off and on, so I'd appreciate a check by someone who knows what a real-world output plot is supposed to look like with this code.  @weaverba137 @dkirkby ?

This is housecleaning for a 21.12 software release for Fuji, but is low priority in the big picture.

### DIFF
--- a/py/desiutil/plots.py
+++ b/py/desiutil/plots.py
@@ -715,10 +715,11 @@ def plot_grid_map(data, ra_edges, dec_edges, cmap='viridis', colorbar=True,
     # Build a 2D array of grid line intersections.
     grid_ra, grid_dec = np.meshgrid(proj_ra, ax.projection_dec(dec_edges))
 
+    ax.grid(False)
     mesh = ax.pcolormesh(grid_ra, grid_dec,
                          data, cmap=cmap, norm=norm, edgecolor='none', lw=0)
 
-    # pcolormesh turns the grid off, turn it back on.
+    # grid turned off for pcolormesh; turn it back on.
     ax.grid(True)
 
     if colorbar:


weaverba137:
@sbailey, there is [an example notebook](https://github.com/desihub/desiutil/blob/master/doc/nb/SkyMapExamples.ipynb) that shows what these plots are supposed to look like.  You may want to run through that, but I can do that too if you don't have time.

sbailey:
@weaverba137 yes, it would be helpful if you could run the notebook and do the checks.  Please also check with a custom kernel that tests with matplotlib 3.5; from the command line one version of that is:
```
base=/global/common/software/desi/users/sjbailey
module use $base/desiconda2/modulefiles

module load desiconda/2.0 
module load desiutil/master
module load desitree/0.5.0
module load QuasarNP/0.1.2
module load specex/0.8.3
module load specter/master
module load desimodel/master
module load desitarget/master
module load specsim/master
module load desispec/master
module load desisim/master
module load fiberassign/master
module load desisurvey/master
module load surveysim/master
module load redrock/master
module load redrock-templates/master
module load prospect/master
module load desimeter/master
module load simqso/master
module load speclite/master
```
(you probably don't need more of those modules past desiutil, but I'm listing what modules are available in that test installation just in case...)  If you have a different setup for testing with alternate package versions, that is fine too.

weaverba137:
I've run through the entire notebook using your matplotlib install.  All the plots looked identical.

sbailey:
Thanks for the cross check; I'll merge now and update changes.rst.

PR #188, opened 2022-08-18
[WIP] Add method for compiling code in a branch

### LINKED ISSUE #186: desiInstall of specex/main and fiberassign/main, which need compilation too
desiInstall supports "in-place" installations of python repos that adds `$PRODUCT_DIR/py` to `$PYTHONPATH` and `$PRODUCT_DIR/bin` to `$PATH` so that any changes to the repo are automatically available without having to do a separate installation step.  Good.

specex and fiberassign, however, are hybrid repos that have python code with compiled extensions.  An in-place install is handy when making changes to any of the python code, but if any of the C++ code changes it still has to be compiled using:
```
python setup.py build_ext --inplace
```

desiInstall doesn't know this, and this pattern doesn't fit any of the build types listed at https://desiutil.readthedocs.io/en/latest/desiInstall.html#determine-build-type .

What's the best way to get desiInstall to know that it needs to run this extra step for these two repos?

A somewhat hacky solution that may not require changing desiInstall is to leverage its special case of looking for an `etc/{productname}_data.sh` script and executing that, e.g. as used by desimodel to get the data from svn.  specex and fiberassign could add their own `etc/*_data.sh` scripts to run `python setup.py build_ext --inplace`, but that is somewhat cryptically using a data-download hook for other purposes.

It might be better to define another hook similar to `etc/*_data.sh`, and if desiInstall detects that it will run it for in-place branch installations, but not for regular installations.  That requires an update to both desiInstall and the specex+fiberassign repos, but it might be more obvious and maintainable in the future.

For context, both specex and fiberassign used to have a Makefile that desiInstall knew to run, but both have migrated to a python-first approach with compiled extensions without a Makefile.  Current master (now main) installations have bootstrapped the `python setup.py build_ext --inplace` upon first installation, after which the desitest nightly update cronjob re-runs that every night after `git pull`.  The point of this ticket is so that the end-user doesn't have to remember to do special steps whenever they make a fresh main installation.

@weaverba137 thoughts?

weaverba137:
This PR closes #186.

@sbailey, I've done one test so far, but I'd like to have more people testing this before merging.

On NERSC/perlmutter, desiutil/main already points to this branch.

### DIFF
--- a/py/desiutil/install.py
+++ b/py/desiutil/install.py
@@ -921,11 +921,16 @@ class DesiInstall(object):
                 if self.options.test:
                     self.log.debug('Test Mode. Skipping compile script.')
                 else:
+                    current_dir = os.getcwd()
+                    self.log.debug("os.chdir('%s')", self.install_dir)
+                    os.chdir(self.install_dir)
                     proc = Popen([compile_script, sys.executable], universal_newlines=True,
                                  stdout=PIPE, stderr=PIPE)
                     out, err = proc.communicate()
                     status = proc.returncode
                     self.log.debug(out)
+                    self.log.debug("os.chdir('%s')", current_dir)
+                    os.chdir(original_dir)
                     if status != 0 and len(err) > 0:
                         message = "Error compiling code: {0}".format(err)
                         self.log.critical(message)
--- a/py/desiutil/install.py
+++ b/py/desiutil/install.py
@@ -909,6 +909,28 @@ class DesiInstall(object):
                     raise DesiInstallException(message)
         return
 
+    def compile_branch(self):
+        """Certain packages need C/C++ code compiled even for a branch install.
+        """
+        if self.is_branch:
+            compile_script = os.path.join(self.install_dir, 'etc',
+                                          '{0}_compile.sh'.format(self.baseproduct))
+            if os.path.exists(compile_script):
+                self.log.debug("Detected compile script: %s.", compile_script)
+                if self.options.test:
+                    self.log.debug('Test Mode. Skipping compile script.')
+                else:
+                    proc = Popen([compile_script, sys.executable], universal_newlines=True,
+                                 stdout=PIPE, stderr=PIPE)
+                    out, err = proc.communicate()
+                    status = proc.returncode
+                    self.log.debug(out)
+                    if status != 0 and len(err) > 0:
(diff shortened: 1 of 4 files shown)


py/desiutil/install.py
```diff
@@ -909,6 +909,28 @@ def get_extra(self):
...
+        """
+        if self.is_branch:
+            compile_script = os.path.join(self.install_dir, 'etc',
+                                          '{0}_compile.sh'.format(self.baseproduct))
+            if os.path.exists(compile_script):
+                self.log.debug("Detected compile script: %s.", compile_script)
+                if self.options.test:
+                    self.log.debug('Test Mode. Skipping compile script.')
+                else:
+                    proc = Popen([compile_script, sys.executable], universal_newlines=True,
```
sbailey, line 923: The example script in desihub/fiberassign#435 includes `cd ${INSTALL_DIR}` so that it moves to the top-level directory before running `python setup.py build_ext --inplace`.  Although that works, it introduces a coupling between package scripts needing to know about environment variables temporarily set by `desiInstall` but not usually set in any other context.  How about handling that within desiInstall itself, e.g.
```python
current_dir = os.getcwd()
os.chdir(self.install_dir)
proc = Popen(...)
os.chdir(current_dir)
```
i.e. we would guarantee that the script is run from the top-level directory and the script itself doesn't have to figure out how to get there (but it could do additional relative `cd` if it needed, e.g. to descend into a subdir for some other compile/config step).

weaverba137: I'm not _opposed_ to this suggestion, but it would create a policy inconsistency, because the _data_ install scripts _are_ allowed to know about `${INSTALL _DIR}`. Just look at the desimodel script.

sbailey: Thanks for pointing out the $INSTALL_DIR usage in desimodel/etc/desimodel_data.sh.  At least in that case the use of $INSTALL_DIR is optional (if it isn't set it just skips that step).

I'm not proposing to go change the data scripts usage here, but at minimum let's not further propagate an $INSTALL_DIR coupling into new scripts in additional repos.

weaverba137:
Code is now updated as requested, as is fiberassign. Please test again.

sbailey (approved):
Works great, thanks!

PR #189, opened 2022-08-18
add gpu_specter to known packages

sbailey:
Small update to add gpu_specter to the list of known packages.

Currently `desiInstall gpu_specter ...` correctly guesses what to do and logs a warning that it is guessing; this update just makes it so that it doesn't have to guess and thus doesn't log a harmless warning.


weaverba137:
This is fine, merging momentarily.

PR #190, opened 2022-08-30
Replace setup.py plugins with stand-alone scripts

### LINKED ISSUE #146: Research replacements for setup.py
As noted in #144, `python setup.py` is being gradually deprecated and replaced by other tools.

In #145, we replaced `python setup.py test` and `python setup.py build_sphinx` with `pytest` and `sphinx-build`, respectively.

In addition, we need to replace:
 * `python setup.py install`
   - This should be replaced by `pip install`, but getting the options correct and compatible with `desiInstall` will take some research.
 * `python setup.py api`
   - Provided by desiutil itself.
   - Generates a doc/api.rst file.
   - Replace with a standalone executable?
 * `python setup.py module_file`
   - Provided by desiutil itself.
   - Generates the module file for the package from the template in the `etc/` directory.
   - Replace with a standalone executable?
 * `python setup.py version`
   - Provided by desiutil itself.
   - Updates `packagename/_version.py`.
   - Two possible replacements:
     * standalone executable
     * [setuptools_scm](https://pypi.org/project/setuptools-scm/)

### LINKED ISSUE #155: Redirect module is tested, but with zero coverage
The tests of the redirect module, though successful, report 0% coverage.  This is strange because even *importing* a module is a test and should result in non-zero coverage.

At minimum, the module should be imported by the test suite, the functions in the module that do not require a script with an independent STDOUT should be tested, and wherever possible, patches and mocks should be used to test the correctness of the code, even if STDOUT can not be redirected.

weaverba137:
This PR:

* Closes #146 
* Closes #155 -- somewhat inadvertently though. In any case the coverage of `redirect.py` is much better now.
* Moves desiutil *itself* in the direction of deprecating the `setup.py` file.
* Cleans up some long-standing style problems so that the style is clean going forward.

The intention is to tag desutil/4.0.0 after this PR, thus marking this as a non-backaward-compatible change.


sbailey:
for help in reviewing this: are any of the changes not backwards compatible at the API level, or is the only non-backwards compatibility to how setup.py is used?  With 47 files changes for mostly style reasons, it could be easy to miss a small but substantive change.

weaverba137:
It is only non-backwards-compatible for `setup.py`.  Even the functions that would have been previously invoked in setup.py are largely unchanged, just rearranged to be called by a `main()` function in a script instead.

weaverba137:
@sbailey, this is the PR that converts the api generation code into a stand-alone script.

weaverba137:
After discussion with @sbailey, we will postpone review & merge until after the `iron` processing run.

sbailey:
Coming back to this PR, post-iron:  will merging this break any of our existing setup.py usage in other packages, or otherwise require changes to procedures?  e.g.
  * python setup.py version
  * python setup.py build_sphinx
  * python setup.py install
  * desiInstall desispec X.Y.Z

Or is this setting the stage for moving away from setup.py, without actually *requiring* us to do so?  If it requires us to make any changes elsewhere, we should clearly itemize that list and make sure that we have reserved the time to do so before merging this.  At this risk of always having "one more thing", I don't want to inadvertently create a distraction for us while also trying to get EDR out the door.

weaverba137:
Yes, as I've said before, this is not a backwards-compatible change, but *only* with regards to how this and other packages use setup.py. Specifically:

* `python setup.py version` --> `desi_update_version`. But I don't recall that we actively use `python setup.py version` extensively.
* `python setup.py build_sphinx` --> We *already* don't do this. We use `sphinx-build` instead.
* `python setup.py install` --> We have already transferred to using `pip install`, especially internally to `desiInstall`.
* `desiInstall desispec X.Y.Z` --> This PR makes *no* changes to `desiInstall`.

Now it's possible that individuals are still typing some of these commands by hand, but most automation has already transitioned.

sbailey:
While focusing on getting EDR out the door, I don't have the cycles to track an optional breaking change to procedures (e.g. reviewing and updating documentation that mentions the old methods).

Reviving an idea that I know we've discussed but I don't recall the conclusion: could we make this a more gentle transition, such that 
```python
from desiutil.setup import DesiTest, DesiVersion, get_version
```
would continue to work, and trying to use them would bring an informative deprecation message telling the user what to run instead?  That would provide the new way of doing things while buying us some time to update docs and setup.py in other packages.

If that isn't straightforward to do, let's postpone merging this again until after EDR is out the door.

weaverba137:
OK, I think we can continue to do that, and I can add deprecation warnings to those functions.  `get_version` should be unaffected by this PR, it's only the `DesiTest` and `DesiVersion` classes that would have gone away.

weaverba137:
Given the above, I recommend that we tag version 3.3.0 after the PR, and reserve desiutil 4.0 for when we finally actually remove deprecated support.

weaverba137:
@sbailey, I can't remember where we left the final decision on this PR.  I restored the functionality you requested previously.  Did you need additional time to test this?

sbailey:
Looks good, thanks.  I double checked that the previous functionality still exists and/or provides useful deprecation messages for what to do instead to help ease the transition.  Please go ahead and merge and tag 3.3.0.  We can then update the procedures documentation and other packages before doing the final removal of the deprecated code here (after EDR).

PR #195, opened 2023-05-23
Update permission settings when installing

### LINKED ISSUE #192: Allow desiInstall to install packages with world-read
Currently `desiInstall` removes world-read permissions when installing DESI packages.  Since, in principle, we want all NERSC users to have access to the DESI software stack to work with public data, we need to default to installing world-read permission, with the non-default option to remove world-read if a package (hypothetically) should remain private.

Modulefiles that are installed alongside packages can be world-readable unconditionally.

### LINKED ISSUE #194: Ensure that very old IERS files do not trigger an error
`freeze_iers()` turns off all downloads related to IERS updates, but in addition, the configuration item `iers_degraded_accuracy` should be set to `'ignore'` or `'warn'` so that old cache files do not trigger an error.

There might be a slight preference for `'warn'` so that we're at least *aware* of how old the cached version is.

@sbailey, do you have a preference for warning or ignore?

weaverba137:
This PR:

* Closes #192.
* Closes #194.
* Adds test coverage to other aspects of `desiInstall`.

Summarizing changes to `desiInstall`:

* Module files will always be installed world-read.
* Software files will be installed world-read, and user,group-read-only, unless:
  - ``--no-world`` is specified, in which case world-read is *disabled*.
  - A branch is installed, including `main`, in which case user-write is *enabled*.
* Internally, `fix_permissions.sh` is no longer used. In fact, no external scripts or unix commands are used to change permissions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,8 @@ MANIFEST
 # Other
 .*.swp
 *~
+.vscode
+.env
 
 # Mac OSX
 .DS_Store
--- a/doc/desiInstall.rst
+++ b/doc/desiInstall.rst
@@ -274,7 +274,8 @@ Configure Module File
 :command:`desiInstall` will scan :envvar:`WORKING_DIR` to determine the details that need
 to be added to the module file.  The final module file will then be written
 into the DESI module directory at NERSC.  If ``--default`` is specified
-on the command line, an appropriate .version file will be created.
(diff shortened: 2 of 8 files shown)


sbailey (approved):
Looks good, thanks.  I empirically tested that the new desiInstall behavior works as advertised (and that it matches what we intend for sharing code installations publicly).

Issue #191, opened 2023-03-23
utilities to add units to tables


sbailey:
One of our standard VAC recommendations is to add units to table columns.  It would be helpful if we had an easy tool to do this to help users do the right thing.  Below are suggested interfaces to implement to streamline adding units to a FITS table post-facto.

## function interface

```python
def add_table_units(filename, extname, colunits, strict=True):
    """
    Add units to table in ``filename`` HDU ``extname``

    Args:
        filename (str): filename of FITS file to modify
        extname (str or int): HDU extension name or number
        colunits (dict): dict with key,value = colname,colunits to update

    Options:
        strict (bool): if True (default), require all columns to exists and units to be valid FITS-format units

    If ``strict`` is True, Raises ValueError if a ``colname`` does not appear in that table
    or if a ``colunits`` isn't a valid FITS unit.
    If an input results in a ValueError, the output file remains unchanged.

    Sets the BUNITnn keyswords in ``filename`` extension ``extname``, modifying the file in-place.
    Other extensions are unchanged.
    """
```

Or maybe split out `strictunits` vs. `strictcolnames` instead of using a single `strict` for both

## command line

Wrap the above function in a command line script to post-facto update the units in a FITS file on disk:

```update_table_units --input/-i FILENAME [--extname/-e EXTNAME] --units COLNAME1=UNITS1 COLNAME2=UNITS2 ...```

--extname is optional only if the file contains a single table in HDU 1.

## doing it right the first time

It would also be helpful if we had cookbook documentation about how to add units in the original code writing the tables when the source data is an astropy Table, a numpy structured array, or a Pandas dataframe.  The spectro pipeline does this with [`desispec.io.util.write_bintable`](https://github.com/desihub/desispec/blob/399ddb6961e285898491b14010f29550645a675f/py/desispec/io/util.py#L185), which is not particularly simple and not ideal as a dependency for science codes writing tables if they don't otherwise have a dependency on desispec.  Do we have simpler recipes to recommend?

## Bonus points: comments

It would also be nice if the same utilities could easily add comments to table columns (like `write_bintable`) so that the 
files are more naturally self-documenting.

geordie666:
An additional idea: [desitarget.io.write_with_units()](https://github.com/desihub/desitarget/blob/main/py/desitarget/io.py#L396-L456) automatically looks up the units for common column names from a [yaml file stored in a data directory](https://github.com/desihub/desitarget/blob/main/py/desitarget/data/units.yaml).

[...]

PR #199, opened 2023-08-30
Add units to a Table

weaverba137:
This PR addresses parts of #191, focussed on adding units to a `astropy.table.Table`.

Here's a simple test snippet:

```python
import os
import importlib.resources as ir
from astropy.table import Table
import desiutil.annotate as da
from desiutil.log import get_logger, DEBUG
da.log = get_logger(DEBUG)
csvFile = os.path.join(str(ir.files('desidatamodel')), 'data', 'column_descriptions.csv')
units, comments = da.csv_units(csvFile)
coadd = '/global/cfs/cdirs/desi/public/edr/spectro/redux/fuji/healpix/sv3/dark/99/9994/coadd-sv3-dark-9994.fits'
t = Table.read(coadd, hdu='FIBERMAP')
tt = da.annotate_table(t, units)  # tt is a copy of t; t is not supposed to be modified in this example.
# Try writing out tt.
```

TO DO:

- [x] Add additional unit tests.
- [ ] Operational testing; extending the snippet above.

### DIFF
--- a/py/desiutil/annotate.py
+++ b/py/desiutil/annotate.py
@@ -11,7 +11,10 @@ import csv
 import os
 import sys
 from argparse import ArgumentParser
+import yaml
 from astropy.io import fits
+from astropy.table import Table, QTable
+from astropy.units import UnitConversionError
 from . import __version__ as desiutilVersion
 from .log import get_logger, DEBUG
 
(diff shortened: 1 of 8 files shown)


py/desiutil/annotate.py
```diff
@@ -0,0 +1,376 @@
...
+from argparse import ArgumentParser
+import yaml
+from astropy.io import fits
+from astropy.table import Table, QTable
+from astropy.units import UnitConversionError
+from . import __version__ as desiutilVersion
+from .log import get_logger, DEBUG
+
+
+log = None
```
sbailey, line 22: This currently requires the user to set `desiutil.annotate.log = get_logger()` after import, which is different from (most?) other desi code which does the `log = get_logger()` within each function, i.e. boilerplate logger setup is the responsibility of the code, not the user.  If you want to support desiutil.annotate using a different loglevel than other code, I suggest changing this to
```python
log = desiutil.log.log
```
so that at least some logger is set by default while still allowing it to be overridden.  Or this could be `log = get_logger()` — although normally running code at import time is discouraged, that's what `desiutil.log.log` is anyway.

py/desiutil/annotate.py
```diff
@@ -0,0 +1,376 @@
...
+    if comment:
+        search = ('comment', 'description')
+    for i, column in enumerate(header):
+        for s in search:
+            if s in column.lower():
+                return i
+    raise IndexError(f"No column matching '{search[0]}' found!")
+
+
+def csv_units(filename):
```
sbailey, line 55: Minor: let's prefix this with a verb like `read_csv_units` or `load_csv_units`.

py/desiutil/annotate.py
```diff
@@ -0,0 +1,376 @@
...
+from astropy.table import Table, QTable
+from astropy.units import UnitConversionError
+from . import __version__ as desiutilVersion
+from .log import get_logger, DEBUG
+
+
+log = None
+
+
+def csv_unit_column(header, comment=False):
```
sbailey, line 25: Thanks for supporting "unit" and "units", "comment" and "comments", and allowing this to be case insensitive.

minor maintainability: I found it confusing that a function called `csv_unit_column` used a boolean switch to search for the comments column name, and that the iterable of column names was called `header` (which implied to me a fits-style key/value header).  Consider updating this to something like
```python
def find_column_name(colnames, prefix='unit'):
    """find case insensitive column name starting with `prefix` from colnames"""
```

py/desiutil/annotate.py
```diff
@@ -0,0 +1,376 @@
...
+                    #
+                    try:
+                        log.debug("t.replace_column('%s', t['%s'].to('%s'))", column, column, units[column])
+                        t.replace_column(column, t[column].to(units[column]))
+                    except UnitConversionError:
+                        log.error("Cannot add or replace unit '%s' to column '%s'!", units[column], column)
+            else:
+                log.info("Not setting blank unit for column '%s'.", column)
+        else:
+            log.info("Column '%s' not present in table.", column)
```
sbailey, line 223: let's downgrade this to a `log.debug`, since the likely common case of using desidatamodel `column_descriptions.csv` will have N>>1 columns not in any given table, which generates a lot of logging cruft.

On the flip side, let's explicitly log if the table has columns that are *not* in `units`.  That's not an error case, but I think is worth logging.  Something like
```python
for colname in t.colnames:
    if colname not in units:
        log.info('Column %s not found in input units table', colname)
```

py/desiutil/annotate.py
```diff
@@ -0,0 +1,376 @@
...
+                    t[column].unit = units[column]
+                except AttributeError:
+                    #
+                    # Can't change .unit if it is already set. Try to convert.
+                    #
+                    try:
+                        log.debug("t.replace_column('%s', t['%s'].to('%s'))", column, column, units[column])
+                        t.replace_column(column, t[column].to(units[column]))
+                    except UnitConversionError:
+                        log.error("Cannot add or replace unit '%s' to column '%s'!", units[column], column)
```
sbailey, line 219: consider adding an equivalent units check to `csv_units` and `yml_units` to catch invalid units at the time they are loaded, before attempting to apply them to a table.

OTOH, it if the input file has invalid units for a column that isn't in the table anyway, it could be considered a feature to not complain until you actually try to use it, so I could see an argument either way.

py/desiutil/annotate.py
```diff
@@ -0,0 +1,376 @@
...
+    """Add annotations to `table`.
+
+    Parameters
+    ----------
+    table : :class:`astropy.table.Table` or :class:`astropy.table.QTable`
+        A data table.
+    units : :class:`dict`
+        Mapping of table columns to units.
+    comments : :class:`dict`, optional
+        Mapping of table columns to comments.
```
sbailey, line 189: It looks like `comments` is a placeholder that doesn't to anything yet.  Let's document that, and consider even raising a `NotImplementedError` if comments is not None.

py/desiutil/annotate.py
```diff
@@ -0,0 +1,376 @@
...
+    parser.add_argument('-u', '--units', action='store', dest='units', metavar='UNITS',
+                        help="UNITS should have the form COLUMN='unit':COLUMN='unit'.")
+    parser.add_argument('-v', '--verbose', dest='verbose', action='store_true',
+                        help='Print extra debugging information.')
+    parser.add_argument('-V', '--version', action='version', version='%(prog)s ' + desiutilVersion)
+    parser.add_argument('-Y', '--yaml', action='store', dest='yaml', metavar='YAML',
+                        help="Read annotations from YAML file.")
+    parser.add_argument('fits', metavar='FITS', help='FITS file to modify.')
+    parser.add_argument('output', metavar='FITS', nargs='?',
+                        help='Write to new FITS file. If --overwrite is specified, this value is ignored.')
```
sbailey, line 322: I'm not sure exactly how metavar is used, but let's give them more descriptive names if possible so that the short form of the help is more understandable, e.g.
```
usage: annotate_fits [-h] [-c COMMENTS] ... INPUT_FITS [OUTPUT_FITS]
```

sbailey (commented):
Thanks for adding this.  I added a number of comments inline related to the user interface and maintainability.  Most of these are minor/optional, but curiously the one I care most about is not making the user set the logger after import since it moves boilerplate responsibility to the user instead of the model; it is a different pattern than most of the rest of our code; and science users using this for a VAC may not even know what a desiutil logger is...

py/desiutil/annotate.py
weaverba137: That's fair, but the ultimate goal of this module was to create a command-line script that would add units to a FITS file, so the *script* would handle setting up the log.  It's only because of the change of focus that a particular user (you) has to do that in the code snippet above. Let me emphasize that the code snippet above is *not* how I envision end-users using `annotate_table()`.  It is only for testing within the context of this PR.

In the *pipeline* context, would `log` be always set up already?

py/desiutil/annotate.py
weaverba137: Will do.

py/desiutil/annotate.py
weaverba137: OK, that's reasonable, but `prefix` has to be a `tuple` not a string, because if you look in more detail, sometimes the comments are called "Description".

py/desiutil/annotate.py
weaverba137: Yes, that's fine.

py/desiutil/annotate.py
weaverba137: Right, but that would come into play when implementing the command-line script that adds units to FITS files, as opposed to a `Table`.

However, I should emphasize here that the `UnitConversionError` exception isn't checking for the validity of the unit *format*. That is catching the much more specific case where *existing* units are being *converted* to other units requested by the user, for example, `nm` --> `Angstrom` or `deg` --> `arcsec`.

Still, I agree that there should be a check on the validity of the unit *format*.

py/desiutil/annotate.py
weaverba137: We may want to remove that entirely, since currently there isn't an established mechanism to add comments to a `Table` in a way that could be written out to a file of any type.

py/desiutil/annotate.py
weaverba137: Well, that example is *exactly* how `metavar` is used, and that's reasonable.

py/desiutil/annotate.py
sbailey: Although the original issue #191 described this as a script to post-facto fix pre-existing VAC files, the utility functions you are setting up here could be very useful for both non-expert VAC authors and the spectro pipeline to add units to tables in the first place so that the files are correct from the start (and correct when they are written the next time without needing another patch).  i.e. let's craft this to make it as easy as possible for users to do the right thing (include units in their files).

> In the pipeline context, would log be always set up already?

In desispec, every function that uses the logger calls `log = get_logger()` internal to the function so that the user can just call the function and it works without any setup.  Nothing in desispec/desitarget/specter/redrock requires additional setup to a module after importing it before it can be used.

py/desiutil/annotate.py
weaverba137: Done in [dab51a4](https://github.com/desihub/desiutil/pull/199/commits/dab51a474c40e19cc3dbbc9c2b8cfd0d4d4e199b).

py/desiutil/annotate.py
weaverba137: Done in [dab51a4](https://github.com/desihub/desiutil/pull/199/commits/dab51a474c40e19cc3dbbc9c2b8cfd0d4d4e199b).

py/desiutil/annotate.py
weaverba137: Done in [dab51a4](https://github.com/desihub/desiutil/pull/199/commits/dab51a474c40e19cc3dbbc9c2b8cfd0d4d4e199b).

py/desiutil/annotate.py
weaverba137: Done in [dab51a4](https://github.com/desihub/desiutil/pull/199/commits/dab51a474c40e19cc3dbbc9c2b8cfd0d4d4e199b).

py/desiutil/annotate.py
weaverba137: OK, I guess how it would work then is if I want to set `DEBUG`, I have to do that in the command-line script *first*, then I can call `log = get_logger()` in each function. That will also involve some changes to the unit tests.

weaverba137: Done in [466159f](https://github.com/desihub/desiutil/pull/199/commits/466159f1c52f091adc2f33d583934ac8be1f6780).

weaverba137:
@sbailey, I suggest moving forward with merging this and splitting off the task of verifying unit format into a separate issue.

There is already unit format verification code in desidatamodel, and I additionally propose moving that code to desiutil so that desiutil doesn't have to depend on desidatamodel.

PR #202, opened 2023-09-22
add support for removing dependencies

sbailey:
This PR updates `desiutil.depend` to add support for removing DEPNAMnn/DEPVERnn dependencies:
  * `removedep(header, name)`: remove a single dependency by name
  * `remove_dependencies(header)`: remove all dependencies

This also required some updates to `getdep` and `iterdep` to not trip on dependencies that had been removed, resulting in non-contiguous DEPNAMnn counting.

This will be used by desihub/desispec#2121 which needs to remove DEPNAM/DEPVER keywords from individual file inputs while stacking them so that they don't get incorrectly merged by `astropy.table.vstack` logic which doesn't know about DEPNAMnn/DEPVERnn pairs.

### DIFF
--- a/py/desiutil/depend.py
+++ b/py/desiutil/depend.py
@@ -131,13 +131,11 @@ def getdep(header, name):
     for i in range(100):
         namekey = 'DEPNAM{:02d}'.format(i)
         verkey = 'DEPVER{:02d}'.format(i)
-        if namekey in header:
-            if header[namekey] == name:
-                return header[verkey]
-        elif i == 0:
-            continue  # ok if DEPNAM00 is missing; continue to DEPNAME01
-        else:
-            raise KeyError('{} version not found'.format(name))
+        if namekey in header and header[namekey] == name:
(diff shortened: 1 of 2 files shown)


weaverba137:
Are you still planning to add `collect_dependencies()` in this PR?

sbailey:
> Are you still planning to add collect_dependencies() in this PR?

No.  With the utility functions added here, the code in desihub/desispec#2101 would become
```
mergedep(data.meta, dependencies)
remove_dependencies(data.meta)
```
without needing a separate `collect_dependencies` function to wrap those (and two lines of "mergedep" + "remove_dependencies" seems clearer to me than wondering what "collect_dependencies" does and having to go read that code to find out...)

weaverba137 (approved):
This all looks fine to me.  Go ahead and merge when ready.

sbailey:
Thanks.  I merged and updated changes.rst in main.

PR #207, opened 2024-05-07
Add bounds checks to radec_to_desiname

### LINKED ISSUE #206: radec_to_desiname should include bounds checks
As mentioned in desihub/desispec#2238, NaN can result in obviously bad values of DESINAME:
```python
>>> import numpy as np
>>> from desiutil.names import radec_to_desiname
>>> radec_to_desiname(np.nan, np.nan)
array(['DESI J-922337203685477.5808-922337203685477.5808'], dtype='<U48')
```
Other bounds should also be checked:
* `NaN`
* `Inf`
* RA not in [0, 360).
* Dec not in [-90, 90].

The proposal is that any value that exceeds these bounds should trigger an exception. We can issue exception messages that are easy to understand, *e.g.*, "Dec values outside of the range [-90, 90] detected." but it should still raise the exception rather than "soldiering on".

@sbailey, @akremin, @schlafly any opinions on the exception, or other bounds that should be checked?

weaverba137:
This PR closes #206.

In addition the test matrix is updated to bring the tests more in line with Jura (hence the oddball branch name).

### DIFF
--- a/py/desiutil/names.py
+++ b/py/desiutil/names.py
@@ -14,19 +14,19 @@ import numpy as np
 def radec_to_desiname(target_ra, target_dec):
     """Convert the right ascension and declination of a DESI target
     into the corresponding "DESINAME" for reference in publications.
-    Length of target_ra and target_dec must be the same if providing an
+    Length of `target_ra` and `target_dec` must be the same if providing an
     array or list. Note that these names are not unique for roughly
     one percent of DESI targets, so also including TARGETID in
     publications is highly recommended for uniqueness.
 
     Parameters
     ----------
-    target_ra: array of :class:`float64`
+    target_ra: array of :class:`~numpy.float64`
         Right ascension in degrees of target object(s). Can be float, double,
-         or array/list of floats or doubles
-    target_dec: array of :class:`float64`
+        or array/list of floats or doubles.
+    target_dec: array of :class:`~numpy.float64`
         Declination in degrees of target object(s). Can be float, double,
-         or array/list of floats or doubles
+        or array/list of floats or doubles.
 
     Returns
     -------
@@ -34,9 +34,23 @@ def radec_to_desiname(target_ra, target_dec):
         The DESI names referring to the input target RA and DEC's. Array is
         the same length as the input arrays.
 
+    Raises
+    ------
+    ValueError
+        If any input values are out of bounds.
     """
     # Convert to numpy array in case inputs are scalars or lists
     target_ra, target_dec = np.atleast_1d(target_ra), np.atleast_1d(target_dec)
+
+    inputs = {'target_ra': target_ra, 'target_dec': target_dec}
+    tests = {'NaN values': np.isnan,
+             'Infinite values': np.isinf,
+             'RA not in range [0, 360)': lambda x: (x < 0) | (x >= 360),
+             'Dec not in range [-90, 90]': lambda x: (x < -90) | (x > 90)}
+    for i in inputs:
+        for t in tests:
+            if (tests[t](inputs[i])).any():
+                raise ValueError(f"{t} detected in {i}!")
 
     # Number of decimal places in final naming convention
     precision = 4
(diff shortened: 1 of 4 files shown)


py/desiutil/names.py
```diff
@@ -43,14 +43,14 @@ def radec_to_desiname(target_ra, target_dec):
...
+             ('Infinite values', np.isinf),
+             ('RA not in range [0, 360)', lambda x: (x < 0) | (x >= 360)),
+             ('Dec not in range [-90, 90]', lambda x: (x < -90) | (x > 90)))
     for i in inputs:
-        for t in tests:
-            if (tests[t](inputs[i])).any():
-                raise ValueError(f"{t} detected in {i}!")
+        for key, check in tests:
+            if (check(inputs[i])).any():
+                raise ValueError(f"{key} detected in {i}!")
```
akremin, line 53: This is very informative but could be made more informative with a `np.where()` call before we raise the error such that we can print the values within the array that are an issue.

I don't feel very strongly about this, but Stephen mentioned it in his comment to the issue so he may feel stronger.

py/desiutil/test/test_names.py
```diff
@@ -56,3 +56,8 @@ def test_radec_to_desiname_bad_values(self):
         with self.assertRaises(ValueError) as e:
             outnames = radec_to_desiname(ras, decs)
         self.assertEqual(str(e.exception), "NaN values detected in target_ra!")
```
akremin, line 58: This and line 63 are fine as-is but if a well-meaning contributor changed the error message to be more informative they could break this test. It would be easy to address at that time, but we could also anticipate that now by only require specific substrings to be in `e`, or leave it as sufficient to just have it raise the correct error.

akremin (commented):
Thank you for the quick fix, Ben. I am happy to accept your changes as-is but I left two things in-line for you to consider.

py/desiutil/names.py
```diff
@@ -42,15 +42,16 @@ def radec_to_desiname(target_ra, target_dec):
...
-             ('Infinite values', np.isinf),
-             ('RA not in range [0, 360)', lambda x: (x < 0) | (x >= 360)),
-             ('Dec not in range [-90, 90]', lambda x: (x < -90) | (x > 90)))
+    base_tests = [('NaN values', np.isnan),
+                  ('Infinite values', np.isinf),]
+    inputs = {'target_ra': {'data': target_ra,
+                            'tests': base_tests + [('RA not in range [0, 360)', lambda x: (x < 0) | (x >= 360))]},
+              'target_dec': {'data': target_dec,
+                             'tests': base_tests + [('Dec not in range [-90, 90]', lambda x: (x < -90) | (x > 90))]}}
     for i in inputs:
```
akremin, line 51: Could we change `i` to `coord` or `ra_or_dec` or something a little more informative? It took me a bit to parse the loops and tuple unpacking, so anything helps for readability.

akremin (commented):
Thanks, my previous comments still remain after this update. You're welcome to ignore them if you wish. I have added one additional comment you can choose to address or not. 

Feel free to accept or reject any of the three and merge if satisfied.

weaverba137:
I will review those soon. I was just very busy debugging the tests yesterday.

py/desiutil/names.py
weaverba137: That's fine. I'll change it to `coord`.

py/desiutil/test/test_names.py
weaverba137: I don't think this is a significant concern. If someone changes the code future, then they need to change the test. There's nothing unexpected about that.

weaverba137:
@sbailey, there is a remaining question about how specific we want to make the error messages, for example not just saying "bad value found", but "25 instances of bad values of target_ra found, for example 375 at index 244 of target_ra".

It wouldn't necessarily be that hard to implement something like that, but why can't users just do that themselves?

weaverba137:
Based on discussion with @sbailey yesterday, I think this can be merged in the current state. I will merge shortly unless there are further comments.

sbailey (approved):
Agreed, ok to merge.

PR #208, opened 2024-06-15
Fix issue #25 of desigal

pentyum:
A more compatible method to read the version of other packages whose version is written in "

### DIFF
--- a/py/desiutil/setup.py
+++ b/py/desiutil/setup.py
@@ -295,9 +295,9 @@ def get_version(productname):
         update_version(productname)
     with open(version_file, "r") as f:
         for line in f.readlines():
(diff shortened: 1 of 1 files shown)


weaverba137:
@pentyum, in order to test this PR, could you please add or modify a unit test so that the new functionality is tested?

pentyum:
> @pentyum, in order to test this PR, could you please add or modify a unit test so that the new functionality is tested?

It is very easy to test this modification. Please see the following picture:
modification_desiutil

weaverba137:
OK, we'll go ahead and merge this then. Thank you for your help with this, it is unlikely we would have noticed this issue otherwise.

PR #209, opened 2024-06-21
Add support for version string corner cases.

weaverba137:
This PR allows a wide variety of valid version strings to be parsed by `desiutil.setup.get_version()`.

While developing this, it was discovered that `desiutil.dust.ext_fitzpatrick()` requires `scipy`, but this requirement had never been explicitly listed. It appears that previously, one or more of astropy, healpy or matplotlib would bring in a dependency on scipy, but more recent versions do not have that dependency, so scipy has to be installed explicitly.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -19,7 +19,7 @@ jobs:
                 os: [ubuntu-latest]
                 python-version: ['3.10']  # There are still issues with Numpy <1.23 and 3.11.
                 numpy-version: ['<1.23', '<1.24', '<2.0']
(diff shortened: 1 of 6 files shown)


sbailey (approved):
Support for corner case version strings + tests looks good.

Commentary on the optional scipy dependency:
If pytest is run in an environment without scipy installed, just one of the dust tests fail with an ImportError, and all of the others run and pass, which seems good.  I had originally suggested that we intercept this case and explicitly skip the test that needed scipy, but a single failing test for an obvious reason (ModuleNotFoundError: No module named 'scipy') is clear and doesn't prevent other tests from being useful, so I like this too.

PR #212, opened 2025-01-10
Preparation for PyPI and other fixes

### LINKED ISSUE #210: annotate_fits warning about overriding units that aren't changing
When running annotate_fits on a file that already has units (e.g. to fill in missing units and/or to update columns), it reports warnings like:
```
WARNING:annotate.py:437:annotate_fits: Overriding units for column 'FLUX_G': 'nanomaggy' -> 'nanomaggy'.
WARNING:annotate.py:437:annotate_fits: Overriding units for column 'FLUX_R': 'nanomaggy' -> 'nanomaggy'.
WARNING:annotate.py:437:annotate_fits: Overriding units for column 'FLUX_Z': 'nanomaggy' -> 'nanomaggy'.
WARNING:annotate.py:437:annotate_fits: Overriding units for column 'FLUX_W1': 'nanomaggy' -> 'nanomaggy'.
WARNING:annotate.py:437:annotate_fits: Overriding units for column 'FLUX_W2': 'nanomaggy' -> 'nanomaggy'.
```
In this case, the overridden unit is identical to the unit that is already there and shouldn't generate a warning.  This should only be a warning if it is overriding it to something different.

I think this is a simple fix, but I'm filing a ticket to remember to do this later.

### LINKED ISSUE #211: odd (wrong?) desiname
@akremin 
```
>>> from desiutil.names import radec_to_desiname
>>> radec_to_desiname(40.669625, -0.013277777)
array(['DESI J040.6696000.-132'], dtype='<U22')
```
I think we've hit a corner-case that the algorithm gets wrong. Haven't checked any catalogs yet; I came across this in the wild.

weaverba137:
This PR:

* Fixes #211, and expands tests to catch the specific cases.
* Fixes #210.
* Expands the GitHub Actions CI matrix to include recent Python, Astropy, Numpy.
* Identifies a previously-hidden *optional* dependency on `speclite`.
* Adds a test to ensure that all scripts in `bin/` are registered in `setup.cfg`.
* Replaces `pkg_resources` with `importlib.resources`.
* Removes `desiutil.setup.DesiTest`. That class was causing some serious problems with recent versions of `setuptools`. Eventually we will want to remove other classes from `desiutil.setup`, but removing just that one is good enough for now.
* Code style clean-up, since lots more people may be seeing the package on PyPI.

However this PR is a WIP, since there are some open issues:

- [x] Since the last time we froze the IERS data in desiutil, the format of the data files provided by [IERS](https://www.iers.org/IERS/EN/Home/home_node.html) -- *i.e.* the organization -- has changed, and Astropy 7 expects this new format. We cannot use the existing `iers_frozen.ecsv` file with Astropy 7. However, we should also no longer *need* to use this file, since Astropy itself provides frozen versions. I think we still need to change the configuration to disable downloads, but I don't think we need to maintain a separate, frozen data file. This needs to be tested.
- [x] [This line](https://github.com/desihub/desiutil/blob/ff61eadeabbb06f3c17459c1a12d9895253f5801/py/desiutil/bitmask.py#L246) throws *a lot* of deprecation warnings. It appears that exists to fix a very, very old problem, that I believe has now been fixed upstream, but in any case, at some point, it will stop working. Please take a look at the comments immediately above that line. We need to do some tests to determine how this actually works now.
- [x] I propose that we tag desiutil 3.5.0 after merging this PR, and then shortly after create a desiutil 4.0.0 which will fully-implement the deprecations listed in the `doc/changes.rst` file.

Finally, I have confirmed with a *test* version of the PyPI repository  that desiutil can be successfully uploaded to PyPI.


weaverba137:
After some testing, I've modified the `.names()` method of bitmasks both avoid lots of warnings and work with both Numpy < 2 and Numpy >= 2.

akremin (approved):
Thank you, Ben. I only reviewed names.py and the corresponding tests. The updates look great and bonus points for the clear comments explaining the various zero-pad choices. I'm happy with the changes and will approve that portion of the PR.

weaverba137:
Thank you @akremin!

py/desiutil/dust.py
```diff
@@ -526,12 +526,12 @@ def __init__(self, fname, scaling):
         self.lam_scal = header['LAM_SCAL']
         self.sign = header['LAM_NSGP']  # north = 1, south = -1
 
-    def ebv(self, l, b, interpolate):
+    def ebv(self, gal_l, gal_b, interpolate):
```
sbailey, line 529: I like `gal_l` and `gal_b` better than just `l`, and `b`, but this is an API change that could break client code.  Please grep desispec, desitarget, desisurvey, and tutorials for uses of this function to make sure that they aren't using `ebv(l=..., b=...)`.  Formally this change would require a 4.0 tag, so let's consider undoing it here since it is stylistic, and save it for the 4.0 changes that aren't API backwards compatible.

py/desiutil/depend.py
```diff
@@ -61,18 +61,14 @@
...
-    'DESI_TARGET', 'DESIMODEL',
-    ]
+possible_dependencies = ['numpy', 'scipy', 'astropy', 'yaml', 'matplotlib',
+                         'requests', 'fitsio', 'h5py', 'mpi4py', 'psycopg2', 'healpy',
+                         'desiutil', 'desispec', 'desitarget', 'desimodel', 'desisim', 'desisurvey',
+                         'specter', 'speclite', 'specsim', 'surveysim', 'redrock', 'desimeter',
+                         'fiberassign', 'gpu_specter',]
+possible_envvars = ['DESI_ROOT', 'DESI_SPECTRO_DATA', 'DESI_SPECTRO_REDUX', 'SPECPROD',
+                    'DESI_SPECTRO_CALIB', 'DESI_BASIS_TEMPLATES',
+                    'DESI_TARGET', 'DESIMODEL',]
```
sbailey, line 71: For the record: Personally I don't prefer this formatting style change due to the extra indentation and editor columns needed.  I'm fine with you owning the style for desiutil and making this update here, but I don't want us to start enforcing this in other desi packages if the original author used the former style.

py/desiutil/test/test_install.py
```diff
@@ -624,26 +623,39 @@ def test_permissions(self, mock_chmod, mock_walk):
...
-                                       (join(self.desiInstall.install_dir, 'lib', 'python3.10', 'site-packages'), ['desiutil-1.2.3.dist-info', 'desiutil'], []),
+                                       (join(self.desiInstall.install_dir, 'lib', 'python3.10', 'site-packages',
+                                             'desiutil-1.2.3.dist-info'), [], ['METADATA', 'LICENSE.rst']),
+                                       (join(self.desiInstall.install_dir, 'lib', 'python3.10', 'site-packages',
+                                             'desiutil', '__pycache__'), [], ['__init__.cpython-3.10.pyc', 'module.cpython-3.10.pyc']),
+                                       (join(self.desiInstall.install_dir, 'lib', 'python3.10', 'site-packages',
+                                             'desiutil'), ['__pycache__'], ['__init__.py', 'module.py']),
+                                       (join(self.desiInstall.install_dir, 'lib', 'python3.10', 'site-packages'),
+                                        ['desiutil-1.2.3.dist-info', 'desiutil'], []),
```
sbailey, line 633: Checking: will these tests work for python versions other than 3.10 (which is what we have at NERSC and appears to be the only thing in the test matrix) or is the test somehow mocking up 3.10 even if running under a different version?  If I'm parsing this correctly, the change here is only formatting so we don't necessarily need to fix it now even if this won't work for python >3.10, but I'll ask now while it's on radar.

sbailey (commented):
Thanks for these updates.  I put in one substantive comment about the dust ebv API change for consideration to postpone until we are making other non-backwards compatible changes for the 4.0 tag.  I suggest that we *not* tackle the iers format change here beyond what you've already done with logging a warning; let's save that for a dedicated PR.

Either way, let's not merge this until Monday so that we aren't making a big update just before the weekend.

py/desiutil/dust.py
weaverba137: OK, I'll revert that.

py/desiutil/test/test_install.py
weaverba137: Ah, I see your point. Currently the test matrix includes 3.10 -- 3.13, and tests are passing, so I think the Python version is simulated in this case, but I'll take a closer look.

py/desiutil/depend.py
weaverba137: That's fine. The other aspect is that desiutil already has so few style problems that it is easy to be strict. When I turned on the indentation enforcement there were less than 10 things to fix, so it is not a drain on productivity.

moustakas (approved):
One quick comment in case it's helpful: in [FastSpecFit](https://github.com/desihub/fastspecfit) I fully deprecated the `setup.cfg` file and ported the equivalent functionality to the `pyproject.toml` file, as recommended by [PEP 621](https://peps.python.org/pep-0621) (see also https://github.com/pypa/setuptools/issues/3683). That's obviously not necessary here but I did want to advertise the effort in case we decide to pursue this change for any of the other DESI packages.

Otherwise, the changes here all look excellent.

py/desiutil/dust.py
weaverba137: Reverted, and #213 created to track this for 4.0.

weaverba137:
@moustakas, thank you, I should take a closer look at that. We might want to consider applying that to a (near) future desiutil 4.0.

py/desiutil/test/test_install.py
weaverba137: The value `'python3.10'` is supplied as a return value to a mock object during testing, and then the value is confirmed in subsequent assert statements, so indeed the version is effectively simulated. However, to prevent confusion, I think I will allow that value to match the version of Python being used to prevent future confusion.

py/desiutil/test/test_install.py
```diff
@@ -35,7 +35,8 @@ class TestInstall(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
-        pass
+        cls.py = f"python{sys.version_info.major:d}.{sys.version_info.minor:d}"
```
sbailey, line 38: Minor: since we also have a directory called "py", I suggest that this variable name be called "pyver" or "pyversion" or similar to disambiguate the meaning when it is used later.

sbailey (approved):

py/desiutil/test/test_install.py
weaverba137: Done

weaverba137:
@sbailey, let's discuss some deployment considerations this afternoon before merging.

weaverba137:
The IERS table issue is moved to #214.

Issue #216, opened 2025-03-11
update install.py to access fetched projects on desi-general


sybenzvi:
To install modules at KPNO, we'll need to handle access to projects fetched to `desi-general`, since direct downloads from GitHub are now blocked. The `--additional-products` flag in install.py can handle project URLs of the form
```
git@desi-general:packagename
```
but internally the URL and branch names aren't handled correctly. The `DesiInstall` class probably needs a new download category to supplement the [github flag](https://github.com/desihub/desiutil/blob/1c0e41a00e41d502c06f61b97a4cb55f0c867084/py/desiutil/install.py#L135) used with GitHub projects.

PR #217, opened 2025-03-11
Enable package downloads from desi-general.

sybenzvi:
Update to the installation module that enables access to desihub packages through an external server at KPNO, addressing issue #216.


weaverba137:
@sybenzvi, I'm not entirely sure if the routines that handle the download already have tests, but I am concerned that the test coverage has decreased slightly. I think at minimum, you could test that the desigeneral *option* can be set and interpreted correctly.

sybenzvi:
If we wanted to add full download tests, we'd need to create a `git` user in the bots with authentication available to `desi-general`, which is maintained by @elliotta and @daqiii. Since that's a pain and you don't insist, let's not do that :). I will add a test checking that the `desigeneral` option is set.

weaverba137:
@sybenzvi is this PR ready for final review?

weaverba137:
@sybenzvi can you please give an update on the status of this PR?

weaverba137:
@sybenzvi can you please give an update on the status of this PR?

PR #219, opened 2025-04-23
Support for mask bits in Numpy 2

### LINKED ISSUE #218: maskbit support for uint32 vs. int64 with numpy/2.x
Currently individual bits of `desiutil.bitmask.BitMask` are `desiutil.bitmask._BitMask` objects which inherit from `int` while adding functionality for `.name`, `.comment`, `.mask` etc.  With numpy<2, this works fine for masking an unsigned 32-bit or a signed 64-bit arrays, but this fails for numpy 2.2.4.

```
import numpy as np
from desispec.maskbits import specmask
print('specmask.BADCOLUMN has type {}'.format(type(specmask.BADCOLUMN)))
print('specmask.BADCOLUMN is an int: {}'.format(isinstance(specmask.BADCOLUMN, int)))
print('Masking a 64-bit mask')
mask64 = np.zeros(10, dtype=np.int64)
mask64 |= specmask.BADCOLUMN
print('Masking an unsigned 32-bit mask')
mask32 = np.zeros(10, dtype=np.uint32)
mask32 |= specmask.BADCOLUMN
```
With numpy/1.22.4 (current NERSC DESI default):
```
specmask.BADCOLUMN has type <class 'desiutil.bitmask._MaskBit'>
specmask.BADCOLUMN is an int: True
Masking a 64-bit mask
Masking an unsigned 32-bit mask
```
With numpy/2.2.4:
```
specmask.BADCOLUMN has type <class 'desiutil.bitmask._MaskBit'>
specmask.BADCOLUMN is an int: True
Masking a 64-bit mask
Masking an unsigned 32-bit mask
---------------------------------------------------------------------------
UFuncTypeError                            Traceback (most recent call last)
Cell In[8], line 10
      8 print('Masking an unsigned 32-bit mask')
      9 mask32 = np.zeros(10, dtype=np.uint32)
---> 10 mask32 |= specmask.BADCOLUMN

UFuncTypeError: Cannot cast ufunc 'bitwise_or' output from dtype('int64') to dtype('uint32') with casting rule 'same_kind'
```

It does work to cask the maskbit to `np.uint32` before masking, but that would be pain for user code to always have to know the right cast before using the mask.
```
mask32 |= np.uint32(specmask.BADCOLUMN)  # OK
```
Note: BADCOLUMN is bit 10 (mask value 1024) so it fits within 32 or 64 bits.

I'm not sure what the best solution is to enable the `BitMask` / `_MaskBit` helper classes to support either 32-bit or 64-bit without the user having to constantly cast the mask when using it.

weaverba137:
This PR fixes #218.

This should be tested with other packages such as desispec and desitarget before merging.

### DIFF
--- a/py/desiutil/bitmask.py
+++ b/py/desiutil/bitmask.py
@@ -114,13 +114,6 @@ class _MaskBit(int):
             else:
                 args.append(input_)
         #
-        # This is boilerplate designed to capture ``out`` keywords.
-        #
-        # try:
-        #     outputs = kwargs['out']
-        # except KeyError:
-        #     outputs = (None,) * ufunc.nout
-        #
         # Identify another argument that we can pass the computation to.
         #
         sup = None
@@ -132,19 +125,6 @@ class _MaskBit(int):
             return NotImplemented
 
         return sup.__array_ufunc__(ufunc, method, *args, **kwargs)
-        # results = sup.__array_ufunc__(ufunc, method, *args, **kwargs)
-        # if results is NotImplemented:
-        #     return NotImplemented
-        #
-        # Deal with multiple outputs.
-        #
-        # if ufunc.nout == 1:
-        #     results = (results,)
-
-        # results = tuple((np.asarray(result) if output is None else output)
-        #                 for result, output in zip(results, outputs))
-
-        # return results[0] if len(results) == 1 else results
 
 
 #  Class to provide mask bit utility functions
--- a/py/desiutil/bitmask.py
+++ b/py/desiutil/bitmask.py
@@ -97,8 +97,7 @@ class _MaskBit(int):
         return ('{0.name:16s} bit {0.bitnum} mask 0x{0.mask:X} - ' +
                 '{0.comment}').format(self)
 
-    # def __repr__(self):
-    #     return "_MaskBit(name='{0.name}', bitnum={0.bitnum:d}, comment='{0.comment}')".format(self)
+    # We do not define __repr__(self) since the superclass has an arguably more useful __repr__().
 
     def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
         """This function makes :class:`~desiutil.bitmask._MaskBit` objects "NumPy-aware",
--- a/py/desiutil/bitmask.py
+++ b/py/desiutil/bitmask.py
@@ -101,8 +101,38 @@ class _MaskBit(int):
     #     return "_MaskBit(name='{0.name}', bitnum={0.bitnum:d}, comment='{0.comment}')".format(self)
 
     def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
-        """When working with Numpy functions and objects, replace the input
-        ``_MaskBit`` object with its ``.mask`` attribute.
+        """This function makes :class:`~desiutil.bitmask._MaskBit` objects "NumPy-aware",
+        so that they can be handled by NumPy `ufunc <https://numpy.org/doc/stable/glossary.html#term-ufunc>`_ methods.
+
+        Parameters
+        ----------
+        ufunc : :class:`numpy.ufunc`
+            The NumPy function, *e.g.* :func:`~numpy.bitwise_or`.
+        method : :class:`str`
+            How `ufunc` was called. This is meant to be passed to another operand.
+        inputs : :class:`list`
+            The positional arguments.
+        kwargs : :class:`dict`
+            The keyword arguments.
+
+        Returns
+        -------
+        array-like
+            This function passes the actual computation to another entry in `inputs`.
+
+        Notes
+        -----
+        * In NumPy 1.x, subclasses of :class:`int` could be used exactly as if the
+          object was a plain Python :class:`int`.
+        * In Numpy 2.x, that is no longer the case, so we must explicitly use the
+          array API to tell NumPy how to handle the subclass.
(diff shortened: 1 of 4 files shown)


py/desiutil/bitmask.py
```diff
@@ -100,6 +100,52 @@ def __str__(self):
     # def __repr__(self):
     #     return "_MaskBit(name='{0.name}', bitnum={0.bitnum:d}, comment='{0.comment}')".format(self)
 
+    def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
+        """When working with Numpy functions and objects, replace the input
+        ``_MaskBit`` object with its ``.mask`` attribute.
+        """
```
sbailey, line 106: To help future maintainers, let's expand these comments to explain more about why this is necessary and provide more context for people who haven't read and understood `__array_ufunc__` documentation.  At minimum reference issue #218 and numpy docs for details; ideally also provide examples of what this function enables / what wouldn't work without it and that this was needed due to changes in numpy 1.x to 2.x.

py/desiutil/bitmask.py
```diff
@@ -100,6 +100,52 @@ def __str__(self):
...
+        #
+        # Deal with multiple outputs.
+        #
+        # if ufunc.nout == 1:
+        #     results = (results,)
+
+        # results = tuple((np.asarray(result) if output is None else output)
+        #                 for result, output in zip(results, outputs))
+
+        # return results[0] if len(results) == 1 else results
```
sbailey, line 147: When we've converged on what is necessary, let's remove the remaining commented out boilerplate before merging.

py/desiutil/test/test_bitmask.py
```diff
@@ -181,3 +180,28 @@ def test_print(self):
...
+                      unsigned_array_16, unsigned_array_32, unsigned_array_64):
+            new_array = array.copy()
+            new_array |= self.ccdmask.COSMIC
+            new_array |= self.ccdmask.HOT
+            self.assertTrue((new_array == (array + 18)).all())
+            self.assertTrue((np.bitwise_and(new_array, self.ccdmask.HOT) != 0).all())
+            self.assertTrue((np.bitwise_and(self.ccdmask.HOT, new_array) != 0).all())
+            out_array = array.copy()
+            np.bitwise_xor(new_array, self.ccdmask.BAD, out=out_array)
+            self.assertTrue((out_array == (array + 19)).all())
```
sbailey, line 203: Let's also add tests that combinations of bits also work prior to the operation with the array (they do work, but this is tricky enough code that I want to check corner cases).  e.g.
```
new_array |= self.ccdmask.mask('DEAD|SATURATED')
new_array |= (self.ccdmask.DEAD | self.ccdmask.SATURATED)
```
Let's also explicitely check the syntax like
```
result = (new_array & self.ccdmask.COSMIC)
```
Under the hood I think that is doing the same thing as `result = np.bitwise_and(new_array, self.ccdmask.COSMIC)` but since we're modifying the behavior of low-level operators, let's make sure.

sbailey (changes requested):
Impressive!  Nice work.  I haven't found any corner cases that don't work.  See inline comments for requests for more documentation for future maintainers, and expanded tests to catch any potential future problems.

py/desiutil/test/test_bitmask.py
weaverba137: Agreed.

py/desiutil/bitmask.py
weaverba137: Clean up done.

weaverba137:
@sbailey, do you recall why `__repr__` was commented out for the `_MaskBit` class?

py/desiutil/bitmask.py
weaverba137: doc string is now expanded.

weaverba137:
Doc tests are failing due to a broken download of matplotlib meta-docs. Let me see if this resolves itself in a while.

weaverba137:
PS, but please take a look at the changes.

sbailey (approved):
Updates look good, thanks.

It looks like you commented out `__repr__` almost 10 years ago in https://github.com/desihub/desiutil/commit/c22cd7dd99180bc63cb8aa208c058f2629fdc7a7 as part of making test examples in the docstring itself.  That's fine with me; *not* overriding `__repr__` allows it to act like an int, e.g. printing "ccdmask.DEAD" prints as "4" instead of "_MaskBit(name='DEAD', bitnum=2, comment='Dead pixel')" which is arguably more informative but less useful.  If you want to remove that commented-out code as part of this PR, that's fine.

weaverba137:
Thanx. I also want to check how `int()` behaves, so don't merge yet.

weaverba137:
@sbailey, I think this would be safe to merge at this point, despite the doc test failures. However, let's make sure the downstream packages are happy first.

sbailey:
> let's make sure the downstream packages are happy first.

I confirm that this fixes the desispec mask-related test failures when running with numpy 2.x, and also works with numpy 1.x.  Looks good!  If that is what you meant by confirming that downstream packages are happy, then go ahead and merge when ready.  Otherwise please confirm what other checks you mean.

weaverba137:
No, that covers what I had in mind. I'll merge in the morning though to see if the doc test error has cleared. Also, I've got plans to get to.

PR #221, opened 2025-04-28
Support for NumPy scalars

### LINKED ISSUE #220: maskbit failure with numpy.bool_
The `_MaskBit.__ufunc_array__` changes in PR #219 to support numpy 2 appear to have broken desitarget tests, which are now failing with
```
py/desitarget/test/test_cuts.py:257: in _test_table_row
    desi, bgs, mws = cuts.apply_cuts(targets[0], tcnames=tc)
py/desitarget/cuts.py:2962: in apply_cuts
    desi_target, bgs_target, mws_target = targcuts.set_target_bits(
...
        # Construct the targetflag bits for DECaLS (i.e. South).
>       desi_target = (lrg_south & res_south) * desi_mask.LRG_SOUTH
E       TypeError: operand type(s) all returned NotImplemented from __array_ufunc__(<ufunc 'multiply'>, '__call__', False, 65536): 'bool_', '_MaskBit'

py/desitarget/cuts.py:2663: TypeError
```
This can be reproduced by running `pytest py/desitarget/test/test_cuts.py` in a desitarget checkout with the latest desiutil.

In this case `(lrg_south & res_south)` evaluates to a `np.bool_`, which triggers the incompatibility with `_MaskBit`; a normal bool is fine:
```
import numpy as np
from desitarget.targetmask import desi_mask
print(f'{type(desi_mask.LRG_SOUTH)=}')
b1 = True
b2 = np.bool_(b1)
print(f'Regular bool {b1 * desi_mask.LRG_SOUTH=}')
print(f'numpy.bool_ {b2 * desi_mask.LRG_SOUTH=}')
```
Results:
```
type(desi_mask.LRG_SOUTH)=<class 'desiutil.bitmask._MaskBit'>
Regular bool b1 * desi_mask.LRG_SOUTH=65536
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[11], line 7
      5 b2 = np.bool_(b1)
      6 print(f'Regular bool {b1 * desi_mask.LRG_SOUTH=}')
----> 7 print(f'numpy.bool_ {b2 * desi_mask.LRG_SOUTH=}')

TypeError: operand type(s) all returned NotImplemented from __array_ufunc__(<ufunc 'multiply'>, '__call__', True, 65536): 'bool_', '_MaskBit'
```

@weaverba137 please investigate an update for `_MaskBit` that would support this case.

@geordie666 heads up, desitarget is broken with desiutil/main and desiutil/3.5.1, but it will works with tag desiutil/3.5.0 and prior.

weaverba137:
This PR fixes #220.

In #219, I think there was a miscommunication:

> I confirm that this fixes the desispec mask-related test failures when running with numpy 2.x, and also works with numpy 1.x.

I assumed that meant that all tests with desitarget, desispec, redrock, and any other package in the nightly desitest suite were succeeding. But that doesn't seem to be consistent with #220. But maybe that meant *only* desispec? Please clarify.

In any case, let's make sure that *every* package is happy with these changes before merging.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -14,7 +14,12 @@ Change Log
 3.5.2 (unreleased)
 ------------------
 
-* No changes yet.
+* Fix operations involving :class:`desiutil.bitmask._MaskBit` and `NumPy scalars`_;
+  remove long-deprecated import of ``setuptools.sandbox`` and associated test code
+  (PR `#221`_).
+
+.. _`#221`: https://github.com/desihub/desiutil/pull/221
+.. _`NumPy scalars`: https://numpy.org/doc/stable/reference/arrays.scalars.html
 
 3.5.1 (2025-04-25)
 ------------------
--- a/py/desiutil/bitmask.py
+++ b/py/desiutil/bitmask.py
@@ -135,11 +135,14 @@ class _MaskBit(int):
         """
         #
         # Filter inputs and replace any _MaskBit with a Python int.
+        # np.generic is the superclass of all NumPy scalars.
         #
         args = []
         for input_ in inputs:
             if isinstance(input_, _MaskBit):
                 args.append(input_.mask)
+            elif isinstance(input_, np.generic):
+                args.append(np.asarray(input_))
             else:
                 args.append(input_)
         #
--- a/py/desiutil/test/test_bitmask.py
+++ b/py/desiutil/test/test_bitmask.py
@@ -224,3 +224,15 @@ class TestBitMask(unittest.TestCase):
         """Test behavior of _MaskBit when explicitly casting to int.
         """
         self.assertEqual(self.ccdmask.HOT.mask, int(self.ccdmask.HOT))
+
+    def test_numpy_scalar(self):
+        """Test behavior of _MaskBit when other operand is a numpy scalar.
+        """
+        for b in (False, True):
+            for t in (np.bool_, np.int16, np.int32, np.int64, np.uint16, np.uint32, np.uint64):
+                result = t(b) * self.ccdmask.COSMIC
+                self.assertEqual(result, int(b) * self.ccdmask.COSMIC)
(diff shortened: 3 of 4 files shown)


weaverba137:
Tests are failing because a long-deprecated feature has been removed from recent versions of `setuptools`.
```
ImportError while importing test module '/home/runner/work/desiutil/desiutil/py/desiutil/test/test_setup.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.11.12/x64/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
py/desiutil/test/test_setup.py:12: in <module>
    from setuptools import sandbox
E   ImportError: cannot import name 'sandbox' from 'setuptools' (/opt/hostedtoolcache/Python/3.11.12/x64/lib/python3.11/site-packages/setuptools/__init__.py)
```

`setuptools.sandbox` is only imported to support `python setup.py test`, which we don't support any longer anyway. Shall we remove this as part of this PR?

weaverba137:
And the import of `setuptools.sandbox` was only in the test suite so it could be removed without changing the actual code.

weaverba137:
After messing around with type casting in numpy 1.x, I now really like numpy 2.x.

weaverba137:
@sbailey, handing this off to you for testing with other packages.

sbailey:
Indeed, previously I had only tested the desispec tests, but not all of the other packages and so missed this one. I'll investigate how to run the nightly tests with desiutil on a different branch than main and report back.

sbailey (approved):
The full suite of nightly tests pass with numpy 1.22.4 and this version of desiutil (except desisim, which fails anyway for unrelated reasons).  Thanks for the fix.

PR #222, opened 2025-07-10
Remove unneeded requirements

weaverba137:
This is a very minor PR that removes references to `urllib3<2` from `doc/rtd-requirements.txt`. `urllib3<2` is being flagged as a security vulnerability by GitHub, even though it is not part of active user code.


weaverba137:
@sbailey, Yes, I checked them most recent RTD build was successful.

PR #224, opened 2025-09-16
Use Astropy-provided IERS data with Astropy >= 7

### LINKED ISSUE #214: Use astropy-supplied data files with freeze_iers()
Astropy now supplies frozen versions of Earth rotation and leap second tables ([astropy-iers-data](https://pypi.org/project/astropy-iers-data/)), which are automatically installed when Astropy is installed.

At the same time, the older, frozen version of the Earth rotation table previously installed by `desiutil.iers.update_iers()` is too old to be used with Astropy 7, since the format of the data has changed, and astropy 7 expects the new format.

For future desiutil 4.0, we propose to abandon the separate table internal to desiutil and use Astropy's tables. However, `freeze_iers()` is probably still needed at some level to turn off any possible downloads of updates.

The current `iers_frozen.ecsv` was added 6 April 2020. Although it was modified in 2021, that modification was to the *metadata* of the file, not the actual data table.

In addition, some operational testing is needed at least to cover these scenarios:

1. If a user doesn't execute `freeze_iers()` triggers a download of IERS data, we assume that version lives in the user's Astropy cache directory. We want to ensure that this cached version will be ignored when `freeze_iers()` *is* executed. In other words, `freeze_iers()` should guarantee that only the official files in `astropy-iers-data` are ever used.
2. Fiberassign reproducibility: we need to thoroughly test fiberassign reproducibility with newer IERS tables. Given that the desiutil IERS data table was updated in April 2020, that basically means the entire survey. We will need to bring in and inform the survey ops team for this.

weaverba137:
This PR closes #214.

Additional testing is needed before merging.

No changes to downstream code should be needed. `desiutil.iers.freeze_iers()` will continue to disable automatic downloads of files. However, with Astropy >= 7, it will no longer attempt to substitute frozen IERS files stored internally for pre-installed files supplied by Astropy via the [astropy-iers-data](https://pypi.org/project/astropy-iers-data/) package.

The internal code of `freeze_iers()` has been rearranged to emphasize the continuing importance of modifying astropy configuration settings over substituting frozen files.


weaverba137:
Note to self: the coverage on `redirect.py` fell significantly.

sbailey:
Initial review: I tested this branch with the current DESI main environment (astropy/6.0.1) and a test environment with astropy/7.1.0 and running from desisurveyops
```
validate_lastnight_fba --night 20220501 --specprod daily   --tempdir $OUTDIR/temp --outroot $OUTDIR
```
The 11 tiles on that night passed the reproducibility check.  I'd like to test the functionality more, but I don't see any issues yet.

weaverba137:
Here's what I've identified so far. Around [Line 43](https://github.com/desihub/desiutil/blob/645e896b92fbf805933ec09e571827421cb9dcdc/py/desiutil/iers.py#L43), `astropy.utils.iers.conf.auto_download = False` is applied *at import time*.

Here's where it gets a bit more speculative: when I clear my Astropy cache, and then run `test_altaz2radec()` the *first* time, https://datacenter.iers.org/data/9/finals2000A.all is downloaded(*). After importing `desiutil.iers`, `auto_download` is `False`, but the import also changes the IERS URL, `astropy.utils.iers.conf.iers_auto_url = 'ftp://cddis.gsfc.nasa.gov/pub/products/iers/finals2000A.all'`.

So the *second* time, Astropy says "do I have a recent IERS file?", so it looks for `ftp://cddis.gsfc.nasa.gov/pub/products/iers/finals2000A.all` in the cache, *not* `https://datacenter.iers.org/data/9/finals2000A.all`. Then it sees it doesn't have that file, but it's not allowed to download it either because `auto_download` is `False`. Hence the exception.

I think those import-time configurations are left over from an issue years ago where the IERS server was offline, so we should just experiment with commenting out those two lines for now.

(*) `https://datacenter.iers.org/data/9/finals2000A.all` is downloaded because `astropy_iers_data.__version__ = "0.2025.8.18.0.40.14"` in your environment and that's more than 30 days old.

weaverba137:
@sbailey another historical note, when `iers_frozen.ecsv` was added to desiutil (in 2020), it was not *created*, it was *moved* from desisurvey (#139). That's why its values only go up to 2018 not 2020.  It was added to desisurvey in 2017 (desihub/desisurvey#39).

sbailey:
From https://github.com/astropy/astropy-iers-data/issues/43, it sounds like the following would be most like the current functionality of freeze_iers with astropy 6.x, i.e. able to work even after importing astropy and freezing to a fixed version of the IERS data regardless of what was installed.
```python
from astropy.utils import iers
iers.IERS_A_FILE = '<path to your file>'
iers.IERS_A_README = '<path to your file>'
iers.IERS_B_FILE = '<path to your file>'
iers.IERS_B_README = '<path to your file>'
iers.IERS_LEAP_SECOND_FILE = '<path to your file>'
```

Although we don't have updated format files going all the way back to our current era of freezing, this sounds pretty good going forward (though we'll still have to fully test the fiberassign reproducibility impact of this transition).  @weaverba137 do you see any blocking factors here?  i.e. copying the files from an old version of astropy-iers-data into desiutil (py/desiutil/data/iers/ or similar), and having freeze_iers use those?

weaverba137:
I don't see any showstopper issues with this. We should identify and document the particular date we want to freeze. I asked on the other thread whether we need to also redefine the README files and didn't see any reply to that question. I just sent a reminder.

weaverba137:
@sbailey at first glance the test failures are all down to tests for exceptions that no longer exist in the simplified code. Testing on *e.g.* fiberassign shouldn't encounter these.

I also want to revise some documentation before merging, but that will probably slip into tomorrow at this point.

weaverba137:
Thank you for the update @sbailey. On that basis, I've moved this PR out of "Draft" status.

sbailey:
Good news: the remaining non-reproducibility with this branch turned out to be an out-of-date desimodel data checkout in the test environment, impacting the list of broken positioners for recently observed tiles. When that was fixed, all tiles passed reproducibility using this branch, with a minor exception of a single potential target on tile 24045 going missing.  This is also the case in main, i.e. current main with numpy/1.x and astropy/6.x also misses this target compared to what was originally run (see desihub/desisurveyops#359).  That was traced to bad-luck rounding at the very edge of the patrol radius for a fiber.

i.e. this PR is good to go. Per usual policy, let's merge on Monday so that we aren't making an optional change going into the weekend.

sbailey (approved):

PR #226, opened 2026-02-11
Demonstrate simplified coverage tests and warning-as-error tests

weaverba137:
This PR demonstrates:

* How to unify coverage tests with the other standard unit tests, thus simplifying the GitHub Actions tests.
* How to set up an Action that runs as a cron job, in this case to catch warnings by converting them to errors.

There are no changes to the user-facing code in this PR.

### DIFF
--- a/.github/workflows/cron-warnings.yml
+++ b/.github/workflows/cron-warnings.yml
@@ -1,0 +1,59 @@
+# Workflow to catch warnings.
+name: Check for warnings (cron)
+
+on:
+    # The push option can be enabled for testing the workflow itself.
+    push:
+        branches:
+            - '*'
+    schedule:
+        # Friday, 22:00 UTC
+        - cron: '0 22 * * 5'
+
+
+jobs:
+    tests:
+        name: Unit tests
+        runs-on: ${{ matrix.os }}
+        strategy:
+            fail-fast: false
+            matrix:
+                include:
+                    # Stable, yet close to bleeding-edge.
+                    - os: ubuntu-latest
+                      python-version: '3.13'
+                      numpy-version: '<3.0'
+                      astropy-version: '<8.0'
+                    - os: ubuntu-latest
+                      python-version: '3.12'
+                      numpy-version: '<2.2'
+                      astropy-version: '<7.1'
+                    - os: ubuntu-latest
+                      python-version: '3.11'
+                      numpy-version: '<2.1'
+                      astropy-version: '<7.1'
+                    - os: ubuntu-latest
+                      python-version: '3.10'
+                      numpy-version: '<2.0'
+                      astropy-version: '<7.0'
+                    # An environment similar to NERSC.
+                    - os: ubuntu-latest
+                      python-version: '3.10'
+                      numpy-version: '<1.23'
+                      astropy-version: '<6.1'
+
+        steps:
+          - name: Checkout code
+            uses: actions/checkout@v6
+          - name: Set up Python ${{ matrix.python-version }}
+            uses: actions/setup-python@v6
+            with:
+              python-version: ${{ matrix.python-version }}
+          - name: Install Python dependencies
+            run: |
+              python -m pip install --upgrade pip setuptools wheel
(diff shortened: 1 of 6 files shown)


weaverba137:
PS, this also adds a short note to `iers.py` about archival data (*i.e.* it's harder to find than I thought).

sbailey (approved):
Thanks for working out the methodology for running coverage tests on only one of the test matrix versions, without having to maintain the installation config separately for coverage.  Does the syntax support if/else?  It took me a minute to notice the `!` vs. not in repeated if-statements like this
```
            - name: Install the package itself
              if: ${{ !fromJSON(matrix.coverage) }}
              run: python -m pip install --editable .[test]
            - name: Install the package itself, with coverage option
              if: ${{ fromJSON(matrix.coverage) }}
              run: python -m pip install --editable .[coverage]
```
I would find it easier to quickly follow the flow of what is happening if it supported syntax like
```
            - name: Install the package itself
              if: ${{ fromJSON(matrix.coverage) }}
              run: python -m pip install --editable .[coverage]
              else:
              run: python -m pip install --editable .[test]
```
or maybe
```
- name: Install the package itself
  run: |
    if [ ${{ matrix.coverage }} == true ]; then
      python -m pip install --editable .[coverage]
    else
      python -m pip install --editable .[test]
    fi
```
If something like that isn't supported, ok.

Cron tests looks good.  For PRs, normally the author of the PR gets email notification if tests fail, but not necessarily others. Do you know who gets notified if the cron test fails and how to configure that?

weaverba137:
I'm pretty sure that `if` is the only construction available, see https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsif. Your second example of effectively creating a mini-bash script may be workable, but it simply moves around the complexity.

weaverba137:
Another option would be to set `matrix.coverage = 'test'` or `matrix.coverage = 'coverage'`:
```yaml
     - name: Install the package itself
       run: python -m pip install --editable .[${{matrix.coverage}}]
```
and if needed:
```yaml
if: ${{ matrix.coverage == 'coverage' }}
```

sbailey:
I like that last option, which would minimize the number of if-statements and would make the remaining ones more obviously different for the test vs. coverage cases.

That leaves the question of who gets notified if the cron test fails, and how to configure that? That doesn't need to be  blocking factor for this PR, but would be good to know.

weaverba137:
> who gets notified if the cron test fails, and how to configure that?

Great question. I'll switch this PR to use the option you mentioned and do some research on that while I'm at it.

weaverba137:
@sbailey see [this document](https://docs.github.com/en/actions/concepts/workflows-and-actions/notifications-for-workflow-runs). Default:

* The person who triggers a GitHub Actions run receives the notification for that action (if any)
* The person who created or last modified a *scheduled* Action (cron job) receives the notification for that workflow.

I believe additional notifications can be received by people who [subscribe](https://docs.github.com/en/subscriptions-and-notifications/concepts/about-notifications), so I suggest you try that with desiutil.

weaverba137:
As I read those documents further, I'm not convinced that it is trivial to alert multiple people to Actions failures. Solutions I've found involve third-party software and configuration details for an external SMTP server.

.github/workflows/cron-warnings.yml
```diff
@@ -0,0 +1,59 @@
...
+name: Check for warnings (cron)
+
+on:
+    # The push option can be enabled for testing the workflow itself.
+    # push:
+    #     branches:
+    #         - '*'
+    schedule:
+        # Friday, 22:00 UTC
+        - cron: '0 22 * * 5'
```
sbailey: Minor: I suggest that we move this to 12:00 UTC on Monday so that we start the US workweek with any alerts rather than getting an alert Friday night that might not get handled until Monday anyway.

sbailey (approved):
Looks good, thanks. I put one minor comment inline for consideration; otherwise merge when ready.

.github/workflows/cron-warnings.yml
weaverba137: Agreed, will change before merging.

PR #230, opened 2026-05-18
add fastspecfit to the list of products which use setuptools_scm

moustakas:
FastSpecFit just switched to using `setuptools_scm` for dynamic versioning, so this PR (+deployment at NERSC) is needed in order for me to install the latest version of the package (3.3.0) at NERSC.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -17,7 +17,10 @@ Change Log
 3.6.2 (unreleased)
 ------------------
 
-* No changes yet.
+* Add `fastspecfit` to the list of packages which use
+  ``setuptools-scm`` to set version strings (PR `#230`_).
+
+.. _`#230`: https://github.com/desihub/desiutil/pull/230
 
 3.6.1 (2026-03-24)
 ------------------
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -17,7 +17,7 @@ Change Log
 3.6.2 (unreleased)
 ------------------
 
-* Add `fastspecfit` to the list of packages which use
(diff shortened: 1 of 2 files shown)


weaverba137:
@moustakas, I want to re-review one of the doc strings that was changed, and then I'll submit an official review.

One important question: do you need this functionality in a desiutil *tag* right away? For example if you're mostly working with packages installed from PyPI instead of the NERSC `main` environment?

doc/changes.rst
```diff
@@ -17,7 +17,10 @@ Change Log
 3.6.2 (unreleased)
 ------------------
 
-* No changes yet.
+* Add `fastspecfit` to the list of packages which use
```
weaverba137: Please use double backticks around fastspecfit. This is just a RST versus MD difference.

weaverba137 (changes requested):
I just suggest one minor change to the change log.

moustakas:
> One important question: do you need this functionality in a desiutil tag right away? For example if you're mostly working with packages installed from PyPI instead of the NERSC main environment?

A new tag would be welcome, but if we have to work off `desiutil/main` it's not the end of the world for the paper with @dirkscholte.

I will need a tag (potentially as early as the end of May or early June) for a new Loa VAC and for the first Matterhorn VAC.

weaverba137:
We can probably get a tag out this week, it's just Mondays can be a bit hectic. @sbailey you're welcome to take a look at this, but I think it's very straightforward.

weaverba137 (approved):
Should be good to go. I can merge later today.

